PR #14, opened 2024-03-25
Wind tunnel recording

ThetaSinner:


ThetaSinner:
Could equally use JSON as the original issue requested but this seems simpler and more compact for now. Happy to change it if we decide we want to be able to read the content back somewhere other than Telegraf.

This PR still needs doc updates and some further testing.

Issue #2, opened 2024-03-11
Wind Tunnel scenarios


ThetaSinner:
This is an epic to cover Wind Tunnel scenarios to be created. It does not need to be complete and can be added to over time

### Regression tests for single conductor performance

- [ ] App installation speed
- [ ] First zome call to a newly installed app
- [ ] Local signal send rate
- [ ] Validation performance with `must_get_agent_activity`
- [ ] Validation performance for simple structures with `must_get_*` other than agent_activity

### Network performance

- [ ] Time to receive required validation receipts
- [ ] Remote call send/receive rate
- [ ] Remote signals send rate
- [ ] Rate of saturation to reach target redundancy with publishing (fast push)

### Network performance with fault injection
- [ ] DHT sync performance with node volatility so that some nodes will come and go during the scenario

### Validation performance

- [ ] Validation rate (in ops/sec)

mattyg:
I added a section for "Regression tests for single conductor performance using wasmer_wamr feature" -- let me know if this issue isn't the best place for that and I should create a standalone issue.

PR #39, opened 2024-05-28
Single conductor scenarios

ThetaSinner:
Single conductor scenarios from #2. Everything else will be going through trycp for now


zomes/signal/coordinator/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+use hdk::prelude::*;
+
+#[hdk_extern]
+fn emit_10k_signals() -> ExternResult<()> {
+    for _ in 0..10_000 {
+        emit_signal("test payload")?;
+    }
```
neonphog, line 7:
```suggestion
    for i in 0..10_000 {
        emit_signal(format!("test payload {i}"))?;
    }
```

Just for debugging? Feel free to ignore this : )

neonphog (approved):
Visual inspection looks awesome to me. Let me know if you'd like me to go more in-depth and actually run the scenarios / look at the influx dashboards. If so I can retract my approval until I do that :+1:

ThetaSinner:
Thanks! I think it'll make sense at some point soon to run this on somebody else's machine. At that point I can walk through running the scenarios and review the dashboards as part of that

PR #46, opened 2024-06-03
Integrate trycp for multi node testing

ThetaSinner:


.github/workflows/test.yaml
```diff
@@ -38,6 +38,17 @@ jobs:
...
+      - uses: actions/cache@v4
+        with:
+          path: |
+            .cargo/bin/
+            .cargo/registry/index/
+            .cargo/registry/cache/
+            .cargo/git/db/
+            target/
+            wasm-target/
+          key: "${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}"
```
neonphog, line 50: Two thoughts:
- this could get large, consider [cargo-sweep](https://crates.io/crates/cargo-sweep) `--stamp` before, then `--file` after.
- I've been pretty happy with [Swatinem/rust-cache](https://github.com/marketplace/actions/rust-cache), but maybe that doesn't integrate with nix as well?

bindings/trycp_client/src/lib.rs
```diff
@@ -0,0 +1,888 @@
...
+                signer,
+                reporter,
+                timeout: Duration::from_secs(30),
+            })
+        }
+
+        pub async fn recv_signal(&mut self) -> Option<Signal> {
+            let mut recv = self.signal_recv.lock().await;
+            recv.recv().await
+        }
```
neonphog, line 61:
```suggestion
        pub async fn recv_signal(&self) -> Option<Signal> {
            let mut recv = self.signal_recv.lock().await;
            recv.recv().await
        }
```

No longer needs to be `mut` if wrapped in a mutex.

bindings/trycp_client/src/lib.rs
```diff
@@ -0,0 +1,888 @@
...
+            let response = self
+                .call_admin(id, AdminRequest::GenerateAgentPubKey, timeout)
+                .await?;
+
+            match response {
+                AdminResponse::AgentPubKeyGenerated(agent_pub_key) => Ok(agent_pub_key),
+                _ => Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    "Unexpected admin response",
+                )),
```
neonphog, line 256:
```suggestion
                _ => Err(io::Error::other("Unexpected admin response")),
```

[other](https://doc.rust-lang.org/std/io/struct.Error.html#method.other) has been stabilized in 1.74

bindings/trycp_client/src/lib.rs
```diff
@@ -0,0 +1,888 @@
...
+            };
+
+            let signed_zome_call = sign_zome_call(zome_call_unsigned, self.signer.clone()).await?;
+
+            let app_request = AppRequest::CallZome(Box::new(signed_zome_call));
+            let response = self.call_app(port, app_request, timeout).await?;
+
+            match response {
+                AppResponse::ZomeCalled(result) => Ok(*result),
+                _ => unreachable!("Unexpected response {:?}", response),
```
neonphog, line 706: Is there a reason this one panics unlike the other apis here?

neonphog (approved):
Woot, looks amazing : )

bindings/trycp_client/src/lib.rs
ThetaSinner: Much tidier, blame copilot! ;) Changed everywhere

bindings/trycp_client/src/lib.rs
ThetaSinner: Good spot, outcompeting clippy there :)

.github/workflows/test.yaml
ThetaSinner: Ahh yes okay, because Holonix moves the cache into the project I'm not that keen to use the `rust-cache`. If we don't maintain that behavior with the new Holonix then I'd be happy to switch. Taken suggestion 1 for now

bindings/trycp_client/src/lib.rs
jost-s, line 1: Apart from the instrumentation, I'd add the methods like `generate_pub_key` etc. to the TryCP client. Not that I mean this for this PR.

scenarios/remote_call/README.md
```diff
@@ -0,0 +1,15 @@
...
+This scenario tests the throughput of `remote_call` operations.
+
+**warning** This is a TryCP-based scenario and needs to be run differently to other scenarios.
+
+### Suggested command
+
+You can run the scenario locally with the following command:
+
+```bash
+RUST_LOG=info cargo run --package remote_call -- --targets targets.yaml --duration 300
```
jost-s, line 14: It doesn't work yet, is that what you meant that it's not complete?

```sh

Running `target/debug/remote_call --targets targets.yaml --duration 300`
#RunId: [TkU1tuVPmBLQF1N3mthta]
[2024-06-04T20:33:40Z INFO  wind_tunnel_runner::run] Running scenario: remote_call
⠁ [---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] [00:00:00 / 00:05:00][2024-06-04T20:33:40Z ERROR wind_tunnel_runner::run] Agent setup failed for agent agent-0: IO error: Connection refused (os error 61)
    
    Caused by:
        Connection refused (os error 61)
⠙ [#>-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] [00:00:02 / 00:05:00][2024-06-04T20:33:42Z ERROR wind_tunnel_runner::run] Agent setup failed for agent agent-1: IO error: No route to host (os error 65)
    
    Caused by:
        No route to host (os error 65)
[2024-06-04T20:33:42Z WARN  wind_tunnel_core::shutdown] Failed to send shutdown signal: SendError(())

Summary of operations
┌──────────────┬─────────────┬─────────────┬─────────────┬──────────────────┬───────────────────┐
│ operation_id │ avg_time_ms │ min_time_ms │ max_time_ms │ total_operations │ total_duration_ms │
└──────────────┴─────────────┴─────────────┴─────────────┴──────────────────┴───────────────────┘`
```

jost-s (approved):

scenarios/remote_call/README.md
neonphog: @jost-s - did you have trycp servers running at the addresses specified in the `targets.yaml` file?

bindings/trycp_client/src/lib.rs
neonphog: They're added here: [https://github.com/holochain/wind-tunnel/pull/46/files#diff-7e33078e2f8938ddfc4bc69066004041d52361604b0f859e566ef20857dfee0cR242](https://github.com/holochain/wind-tunnel/pull/46/files#diff-7e33078e2f8938ddfc4bc69066004041d52361604b0f859e566ef20857dfee0cR242) is that not what you meant @jost-s ?

scenarios/remote_call/README.md
jost-s: I did not. That could be added to this readme.

bindings/trycp_client/src/lib.rs
jost-s: Sorry, the link just opens the PR on code diff tab. What are you referring to?
I mean that instead of keeping wrappers like
```rust
pub async fn download_dna(
  &self,
  url: String,
  timeout: Option<Duration>,
) -> io::Result<String> {
  let response = self
    .trycp_client
    .request(
      Request::DownloadDna { url },
      timeout.unwrap_or(self.timeout),
    )
    .await?;
  read_response(response)
}
```
here I'd move them to the TryCP client package and then call them here
```rust
trycp_client.download_dna()
```

scenarios/remote_call/README.md
jost-s: Is this what's supposed to happen?

```sh
Running `target/debug/remote_call --targets targets.yaml --duration 300`
#RunId: [sb4yG7VP6jLxTKEDjS8vw]
[2024-06-04T20:55:34Z INFO  wind_tunnel_runner::run] Running scenario: remote_call
⠤ [###########################>-----------------------------------------------------------------------------------] [00:01:14 / 00:05:00][2024-06-04T20:56:49Z ERROR wind_tunnel_runner::run] Agent setup failed for agent agent-1: IO error: Operation timed out (os error 60)
    
    Caused by:
        Operation timed out (os error 60)
⠲ [##############################################################################################################>] [00:04:59 / 00:05:00][2024-06-04T21:00:34Z WARN  wind_tunnel_core::shutdown] Failed to send shutdown signal: SendError(())

Summary of operations
┌────────────────────────┬─────────────┬─────────────┬─────────────┬──────────────────┬───────────────────┐
│ operation_id           │ avg_time_ms │ min_time_ms │ max_time_ms │ total_operations │ total_duration_ms │
├────────────────────────┼─────────────┼─────────────┼─────────────┼──────────────────┼───────────────────┤
│ trycp_configure_player │ 533.44      │ 508.32      │ 699.98      │ 561              │ 299259.78         │
├────────────────────────┼─────────────┼─────────────┼─────────────┼──────────────────┼───────────────────┤
│ trycp_reset            │ 0.41        │ 0.31        │ 1.06        │ 562              │ 230.56            │
└────────────────────────┴─────────────┴─────────────┴─────────────┴──────────────────┴───────────────────┘
```

bindings/trycp_client/src/lib.rs
neonphog: We discussed that as an alternative, but it seemed weird to me that would require trycp_client in the tryorama repo to specify an exact version of holochain_conductor_api. Even if whatever version is pinned there may technically work across multiple holochain binary versions, it still restricts it more than the trycp protocol technically does, which will work across all holochain binary until we change something about the websocket interface (like the allow-list thing).

scenarios/remote_call/README.md
neonphog: Looks right to me :+1:

bindings/trycp_client/src/lib.rs
jost-s: I see. Is that in effect not the same since we depend here on a version of the client and the client depends on a version of the trycp_api which defines the interface according to the conductor api?

neonphog: Since the trycp_api only passes AdminRequest/Responses and AppRequest/Responses as encoded bytes, it does not bind us to a specific holochain_conductor_api version. Is that what you were asking?

jost-s: Ah yes, sorry, my oversight, of course, with just bytes as the message it makes all the difference. Cool

neonphog (approved):

scenarios/remote_call/README.md
ThetaSinner: Yep, that looks great.

The local IP in the target.yaml is on my network so yes it won't work for anybody else. I intend to replace the contents of this list with HP addresses.

Once that's done though, it will still be normal for some addresses to be unavailable. So you will see the same error that you're seeing here. That shouldn't and doesn't terminate the scenario, just the one agent stops.

bindings/trycp_client/src/lib.rs
ThetaSinner: Happy to move this client code at some point. I'm in about the same place as this conversation - I'm not entirely clear where the code belongs. Probably not in the instrumentation wrapper though! :)

bindings/client/src/error.rs
```diff
@@ -9,10 +10,11 @@ use std::io::ErrorKind;
 pub fn handle_api_err(err: ConductorApiError) -> anyhow::Error {
     match err {
         // Handle websocket closed errors by shutting down the process, as this is a fatal error
+        // for this agent.
         ConductorApiError::WebsocketError(e)
             if e.kind() == ErrorKind::Other && e.to_string() == "ConnectionClosed" =>
         {
-            panic!("Conductor API connection closed unexpectedly: {:?}", e)
+            AgentBailError::default().into()
```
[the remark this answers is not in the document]
ThetaSinner: Normally:
- an error in the agent setup immediately stops that agent
- all errors are just logged during agent behaviour so that it can keep running
- errors in teardown are just logged because this is just a best effort cleanup

I'm happy to rename this error, but its purpose is to allow the agent to break out of its behaviour loop. If a HP goes down during a test we want to exit the loop, not keep spamming errors and wasting resources on the control node.

.github/workflows/test.yaml
```diff
@@ -144,6 +160,22 @@ jobs:
 
           pkill --echo hc && pkill --echo holochain && pkill --echo lair-keystore
 
+      - name: Smoke test - remote_call
+        run: |
+          # Start a sandbox conductor and run it in the background
+          nix develop --command bash -c "hc s clean && echo "1234" | hc s --piped create && echo "1234" | hc s --piped -f 8888 run &"
```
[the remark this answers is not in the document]
ThetaSinner: Done

bindings/client/src/error.rs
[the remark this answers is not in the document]
jost-s: I think that's ideal and was jokingly asking about some secret agent whose name must be Bail.

jost-s (dismissed):
💯

PR #51, opened 2024-06-13
Remote call rate

ThetaSinner:


conductor-config.yaml
```diff
@@ -0,0 +1,6 @@
+network:
+  network_type: quic_bootstrap
+  transport_pool:
+    - type: webrtc
+      signal_url: "wss://signal.holo.host"
+  bootstrap_service: "https://bootstrap.holo.host"
```
neonphog, line 6: We should probably write a story to move away from production services for performance testing.

scenarios/remote_call_rate/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+                None => {
+                    // No more agents available to call, get a new list.
+                    // This is also the initial condition.
+                    Ok(client
+                        .agent_info(agent_id, None, None)
+                        .await?
+                        .into_iter()
+                        .map(|info| AgentPubKey::from_raw_36(info.agent.0.clone()))
+                        .filter(|k| k != cell_id.agent_pubkey()) // Don't call ourselves!
+                        .collect::<Vec<_>>())
```
neonphog, line 74: `.shuffle()`?

bindings/trycp_runner/src/common.rs
```diff
@@ -45,6 +56,253 @@ pub fn connect_trycp_client<SV: UserValuesConstraint>(
...
+where
+    SV: UserValuesConstraint,
+{
+    let client = ctx.get().trycp_client();
+    let agent_id = ctx.agent_id().to_string();
+
+    ctx.runner_context()
+        .executor()
+        .execute_in_place(async move {
+            client.shutdown(agent_id.clone(), None, None).await?;
```
neonphog, line 299: I know we run reset as part of the startup, but do we want to run it as part of shutdown too (maybe with `let _ = `) so we don't leave the databases of previous tests laying around?

bindings/trycp_runner/src/common.rs
```diff
@@ -45,6 +56,253 @@ pub fn connect_trycp_client<SV: UserValuesConstraint>(
...
+                .ok_or(anyhow::anyhow!("Cell not found"))?
+            {
+                CellInfo::Provisioned(pc) => pc.cell_id.clone(),
+                _ => anyhow::bail!("Cell not provisioned"),
+            };
+            log::debug!("Got cell id: {:}", cell_id);
+
+            let credentials = client
+                .authorize_signing_credentials(
+                    agent_id.clone(),
```
neonphog, line 182: Originally I was confused thinking agent_id here was the agent pubkey, but no, this is an indicator to the multi client which sub-client we're talking to? I don't know if a different variable name would be clearer.

Also, is it true that since we set up the init with unrestricted access, this `authorize_signing_credentials` isn't strictly necessary... except that the calls need to be signed by *something*?

neonphog (dismissed):
This looks amazing! :tada:

conductor-config.yaml
ThetaSinner: Yes, absolutely - https://github.com/holochain/wind-tunnel/issues/52

scenarios/remote_call_rate/src/main.rs
ThetaSinner: Yes, that makes sense. It's possible that because we start peers with some ordering, that they discover each other with some ordering and end up with several making calls to each other at the same time - evening that out is a good idea

bindings/trycp_runner/src/common.rs
ThetaSinner: Yep, particularly for tests that create more data than this one does, that's probably a good pattern to put in place.

Added this to the agent tear down and handled the error at that level

bindings/trycp_runner/src/common.rs
ThetaSinner: Correct, it's nothing to do with the agent pubkey - I've renamed it to `agent_name` which doesn't imply a unique identifier and it hopefully clearer? What that field is for has documentation on the getter function if you're working in the code but I agree it's confusing once it's being reviewed or has been copied into a variable.

Correct that we have unrestricted zome call access but it's the zome calls from the scenario to the conductor that we're setting up signing credentials for. It's effectively pointless when we have access to the admin websocket anyway because we already have complete control over the conductor... but that's what the API requires. Hopefully now that it's hidden in a common function, scenario authors won't have to worry too much about that.

scenarios/remote_call_rate/src/main.rs
```diff
@@ -127,6 +127,11 @@ fn agent_teardown(
 ) -> HookResult {
     shutdown_remote(ctx)?;
     disconnect_trycp_client(ctx)?;
+
+    // Best effort to remove data and cleanup.
+    // You should comment out this line if you want to examine the result of the scenario run!
+    let _ = reset_trycp_remote(ctx);
```
neonphog, line 133: Does this work after we disconnect the trycp_client the line above?

ThetaSinner: Definitely not :)

neonphog (approved):
woot :+1:

PR #60, opened 2024-07-03
Two party countersigning

ThetaSinner:


ThetaSinner:
image

ThetaSinner:
It looks like the session initiation ends up quite uneven. Both try to initiate a good number of sessions but one agent seems to lose out and get lots of failures. I'm slightly perplexed about why that is when there should be enough agents that conflicts are relatively rare. The sessions are pretty fast so if one initiator times out then there's a possibility that they'll run into the other initiator right after. The session timeout is reasonable high compared with the success path so even a small number of collisions will really set you back.

scenarios/two_party_countersigning/README.md
```diff
@@ -0,0 +1,36 @@
+## remote_call
+
+### Description
+
+This scenario tests the throughput of `remote_call` operations.
```
neonphog, line 5:
```suggestion
## two_party_countersigning

### Description

This scenario tests the performance of `countersigning` operations.
```

zomes/countersigning/coordinator/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+        e => {
+            return Err(wasm_error!(WasmErrorInner::Guest(format!(
+                "Unexpected response: {:?}",
+                e
+            ))))
+        }
+    };
+
+    // Let the other party know about the request
+    send_remote_signal(
```
neonphog, line 119: since remote signals are unreliable, and we only have one remote target, would it make more sense for this to be a call_remote?

scenarios/two_party_countersigning/src/main.rs
```diff
@@ -0,0 +1,390 @@
...
+                                "countersigning",
+                                "list_participants",
+                                (),
+                                None,
+                            )
+                            .await
+                            .context("Failed to list participants")?;
+                        let mut new_peer_list = response.decode::<Vec<AgentPubKey>>().map_err(|e| anyhow::anyhow!("Decoding failure: {:?}", e))?;
+                        new_peer_list.shuffle(&mut thread_rng());
+                        Ok(new_peer_list)
```
neonphog, line 137: Just checking my understanding: while we're still setting up waiting for participant links to be published to us, this will effectively be a busy loop running the list_participants zome call as fast as possible?

zomes/countersigning/coordinator/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+        return Err(wasm_error!(WasmErrorInner::Guest(
+            "Only chocolate is accepted".to_string()
+        )));
+    }
+
+    // Accept the request and send the acceptance back to the requester
+    let my_accept = accept_countersigning_preflight_request(request)?;
+    match my_accept {
+        PreflightRequestAcceptance::Accepted(response) => {
+            send_remote_signal(Signals::Response(response.clone()), vec![initiating_agent])?;
```
neonphog, line 146: ditto, do you think this should actually be a call_remote?

scenarios/two_party_countersigning/README.md
```diff
@@ -0,0 +1,36 @@
...
+immediately see one peer after app installation.
+
+This configuration defaults to 2 peers.
+
+### Suggested command
+
+You can run the scenario locally with the following command:
+
+```bash
+RUST_LOG=OFF MIN_PEERS=5 cargo run --package two_party_countersigning -- --targets targets-ci.yaml --behaviour initiate:2 --behaviour participate:3 --instances-per-target 5 --duration 300 --reporter influx-file
```
neonphog, line 27: I wonder, in light of your comment about conflicts, since (to my understanding) that's not the path we're trying to get performance data about in the test, if we'd rather do some string processing and just have some dedicated groups eliminating the possibility of conflict:

```
--behaviour init-a:1 --behaviour part-a: 3 --behaviour init-b:1 --behaviour part-b: 3, etc
```

scenarios/two_party_countersigning/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+This assumes that `trycp_server` is running. See the script `scripts/trycp.sh` and run with `start_trycp`.
+
+To run the scenario against the current target list, you can run:
+
+```bash
+RUST_LOG=info MIN_PEERS=40 cargo run --package two_party_countersigning -- --targets targets.yaml --behaviour initiate:1 --behaviour participate:1 --duration 500
```
neonphog, line 35: mmmm... I wrote my above comment assuming I was talking about the trycp version, but seeing this line, I think we might have to come up with some clever way to distribute behaviors proportionally to the available target list....

neonphog (approved):
Nice :+1: - my comments can be follow-on work if you'd like to get this in first.

.github/workflows/test.yaml
```diff
@@ -164,8 +164,19 @@ jobs:
           # Start a TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && start_trycp"
 
-          # Run the scenario for 30 seconds
-          RUST_LOG=info nix run .#remote_call_rate -- --targets targets-ci.yaml --instances-per-target 2 --duration 30 --no-progress
+          # Run the scenario for 10 seconds
+          RUST_LOG=warn MIN_PEERS=2 nix run .#remote_call_rate -- --targets targets-ci.yaml --instances-per-target 2 --duration 10 --no-progress
```
jost-s, line 168: Is 30 seconds too long?

bindings/trycp_runner/src/common.rs
```diff
@@ -47,7 +47,7 @@ pub fn connect_trycp_client<SV: UserValuesConstraint>(
 
     let client = ctx.runner_context().executor().execute_in_place({
         let signer = signer.clone();
-        async move { Ok(TryCPClient::connect(target, signer.clone(), reporter).await?) }
+        async move { TryCPClient::connect(target, signer.clone(), reporter).await.with_context(|| format!("Could not connect TryCP client, is there a server running and reachable at {}?", target)) }
```
jost-s, line 50: "Evidently not" 😉

zomes/countersigning/coordinator/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+        .map(|link| link.target.clone().try_into().unwrap())
+        .collect())
+}
+
+#[hdk_extern]
+fn start_two_party(with_other: AgentPubKey) -> ExternResult<PreflightResponse> {
+    let my_agent_info = agent_info()?;
+
+    let entry = ImportantAgreement {
+        best_ice_cream_flavour: "strawberry".to_string(),
```
jost-s, line 74: Clearly it's chocolate

scenarios/two_party_countersigning/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+You can run the scenario locally with the following command:
+
+```bash
+RUST_LOG=OFF MIN_PEERS=5 cargo run --package two_party_countersigning -- --targets targets-ci.yaml --behaviour initiate:2 --behaviour participate:3 --instances-per-target 5 --duration 300 --reporter influx-file
+```
+
+This assumes that `trycp_server` is running. See the script `scripts/trycp.sh` and run with `start_trycp`.
+
+To run the scenario against the current target list, you can run:
```
jost-s, line 32: What is the current target list?

scenarios/two_party_countersigning/src/main.rs
ThetaSinner: It will indeed, added a sleep to slow it down when no agents are found.

scenarios/two_party_countersigning/README.md
ThetaSinner: The trycp runner receives the CLI input before it gets passed to the wind tunnel runner, we are free to pre-process the args any way we want. The `instances-per-target` flag is an example of that. Wind Tunnel knows nothing about it, but it tweaks the wind tunnel behaviour by allowing multiple conductors to be launched by one trycp server.

scenarios/two_party_countersigning/README.md
ThetaSinner: We could achieve partitioning within the test if we want to, if we expose the number of peers assigned to a behaviour then we can use `%`-logic to partition the targets by their index. I.e. rather than just advertising "I am a participant", they could advertise "I am participant X" and the initiators could select a subset from that by their own index.

I'm not keen to do this though. It effectively becomes equivalent to "--behaviour init:1 --behaviour part:5 --targets 6` with some  noise from syncing ops. At that point the network, or subgraph of the network, is limited by how fast one agent can execute countersigning sessions. At that point, because the accept is much faster than the initiate we may as well just test with two nodes if we want to measure purely throughput. I think that is a valid way to run this test that we could run a report on.

The more general case though is to allow conflicts to happen. On a real network you wouldn't necessarily know that somebody you want to do countersigning with is already doing a session with somebody else. All I'm really worried about with the conflicts is a system issue. If we see the same behaviour on the HPs that I'm seeing locally then something is wrong with the test

ThetaSinner: I had a though overnight about what that might be. The accepting agents must be listening again before their chain is unlocked. This might be fixable by waiting for their chain to be unlocked before resuming listening to signals. It shouldn't backlog signals more than the number of initiating agents because they will always wait for a response. So doing that should lower the throughput and get rid of any errors... I think

.github/workflows/test.yaml
ThetaSinner: I think so. Unless we assert the results, this is just about getting the code to execute at least once. it doesn't need to run for long.

I'm thinking about adding a 'max iterations' or something like that to the CLI flags for the runner. That way we could set the timeout much higher and limit how many times the behaviour runs to something like 3 times. That would eliminate guessing on CI and ensure that we always test the agent behaviour even when things run slowly. It would also be really useful for debugging. When something goes wrong with a scenario it can really spam errors. If we could limit the number of iterations then it would be easier to stop the scenario and look at the first few errors.

bindings/trycp_runner/src/common.rs
ThetaSinner: Haha, just trying to be helpful. Connection refused isn't very meaningful. 

I do this regularly, so I have to be nice to myself in the messages ;)

zomes/countersigning/coordinator/src/lib.rs
ThetaSinner: I went down this route of creating some fun test data, then really struggled to pick. I wanted chocolate, banana, mint choc-chip and several more.... I really like ice cream :)

scenarios/two_party_countersigning/README.md
ThetaSinner: Not sure how to describe it best here, but it'd the be `targets.yaml` in the root of the project that contains a list of HPs that are currently assigned for this purpose.

zomes/countersigning/coordinator/src/lib.rs
ThetaSinner: This is a good point. We'll probably have to use remote signals to keep the M of N case performant but `call_remote` is a better option here.

ThetaSinner: I'll leave the signal setup in place to be used later but switch the init/accept zome functions to use `call_remote`

zomes/countersigning/coordinator/src/lib.rs
ThetaSinner: Yes, done

ThetaSinner:
Right okay, countersigning emits a signal when it is finished, so you can listen for that to prevent overlap. I've added this and all the errors about locked chains go away. That's great! Unfortunately now I get a bunch of errors about missing DHT dependencies... I really don't understand how this works under the hood then. I guess you actually need quite a bit of synchronisation in the coordination for this to make it work reliably?

scenarios/two_party_countersigning/README.md
ThetaSinner: Yes, that was part of the problem. Digging into Holochain code I found that there is a signal emitted when the session completes. So I've solved one problem with that. Now I have another problem which is that another kind of error was being hidden by the first

ThetaSinner:
Okay, after several fixes and attempted fixes, this still behaves quite poorly. At this point I'm straying into debugging countersigning. That is absolutely not the point of this work, so I think it's time to pause.

I tried running against the Holoports but that doesn't work well with assigned behaviours because you end up with required behaviours assigned to offline nodes... I need to do some kind of pre-flight filter to give the test a reasonable chance of running

scenarios/two_party_countersigning/README.md
neonphog: Awesome, yeah I certainly don't mind having the conflicting sessions in there as long as we can distinguish them in the metrics :+1:

ThetaSinner: Fair point, the error metrics are not very informative at the moment because there are so many possible exit points and I'm really not clear which ones we should care about. Plus the error messages that come back contain far too many strings to be easily processable. I think this will need work when we come to reporting on this. For now all you can really tell is how many sessions were attempted, how many succeeded and how long the successful sessions took. I think that's a useful starting point for now, definitely not reaching what I'd think of as ideal.

PR #77, opened 2024-08-16
trycp version of write_validated scenario

neonphog:
wind-tunnel-write-validated-dashboard-cell

This is essentially a copy of the `write_validated` scenario, but for trycp. Let me know if you were imagining anything in addition to this!


scenarios/trycp_write_validated/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+
+            reporter.add_custom(
+                ReportMetric::new("create_sample_entry_time")
+                    .with_field("value", start.elapsed().as_secs_f64()),
+            );
+
+            let start = Instant::now();
+
+            let _: ActionHash = client
+                .call_zome(
```
ThetaSinner, line 81: The original wind tunnel runner had a `call_zome` common helper in `bindings/runner/common.rs` that does the decoding for you. I wonder if we should have one for the trycp version that grabs the information it needs from the context and does the call with encoding/decoding so that this become the 'raw' option and we can use the helper by default.

Doesn't have to be done on this PR and it's easy to swap one for the other if it's done later :)

ThetaSinner:
This looks good, I can't think of any reason to add more than this. We just want to be able to run the scenario we had already but with TryCP! :)

I do see some differences in the chart to other charts, so that's my inconsistency there!
- The color scheme is set to color blind friendly, I just think that's a good default after I've had many people complain about UI work I've done because they can't tell what's what :) Optional change!
- More important, there's no unit on the y-axis. That should be in `s` so that it will render as `ms`.

Would you be up for tweaking the dashboard, then exporting it and adding the changes to the PR?

---

This scenario should be added to the CI smoke tests in the build pipeline - there are TryCP examples already, so you can pretty much copy one of those and put the command from the README for this scenario into the middle bit where it runs the scenario.

---

One other question, have you run this with a few more nodes just to make sure it's behaving as expected? Something like

```
RUST_LOG=info CONDUCTOR_CONFIG="CI" TRYCP_RUST_LOG="info" MIN_PEERS=10 cargo run --package trycp_write_validated -- --targets targets-ci.yaml --instances-per-target 10 --duration 300
```

The code all looks good to me, but my experience so far is that it's easy for problems to hide with 2-3 nodes and it's worth running for a few minutes with more nodes. If that works then it's generally the case that any problem that appear on the Holoports are to do with live environment conditions rather than problems with the scenario

neonphog:
> it's worth running for a few minutes with more nodes

Seems to work for me. I do get a couple:

```
[2024-08-20T16:07:33Z ERROR wind_tunnel_runner::run] Agent behaviour [default] failed for agent agent-6: call failure: Custom { kind: Other, error: "InternalError(\"Source chain error: deadline has elapsed\")" }
```

But I assume that's just my laptop being overloaded, and the scenario seems to keep going and reporting.

neonphog:
> I do see some differences in the chart to other charts

Okay, added a write-validated dashboard: https://github.com/holochain/wind-tunnel/pull/77/commits/25f04c65c9991a16d5c4ac98a7bb3418a89ff469

wind-tunnel-write-validated-dashboard-cell

neonphog:
> This scenario should be added to the CI smoke tests in the build pipeline

attempt here: https://github.com/holochain/wind-tunnel/pull/77/commits/be6b6ef7da543963548260c29767d4cac2cb87e5

We'll see if it works : )

scenarios/trycp_write_validated/src/main.rs
neonphog: Good idea, this cleaned up the scenario a bunch: https://github.com/holochain/wind-tunnel/pull/77/commits/a7ad0d9ce0da34b331ddee5c59f9d6e621e75bbb

ThetaSinner: Very nice, yeah if you can make the common functions do the nasty task spawn then the scenario ends up much cleaner

ThetaSinner (approved):

PR #83, opened 2024-08-29
Validation Receipts Scenario

### LINKED ISSUE #44: Wind Tunnel Scenario: Time to receive required validation receipts
This is a placeholder split off from https://github.com/holochain/wind-tunnel/issues/28

neonphog:
RESOLVES #44 

validation_receipt_timing

Hehe ^^ I thought this graph was a little too awesome. It turns out it was breaking out of the loop early if it hadn't received any receipts at all yet. With that fixed, I'm seeing much more expected 3 to 6 second timing on this.

---

Graph now split by op type in "wait_for_complete" mode:

validation_receipts_wait_complete

---

New graph in "no_wait_for_complete" mode... much weirder graph that stops getting data part way through. I assume that is because a bunch of the actions *never* actually complete, especially as it goes on. But that can be something to figure out in another story:

validation_receipts_no_wait_complete


conductor-config-ci.yaml
```diff
@@ -1,3 +1,6 @@
+dpki:
+  device_seed_lair_tag: ci
+  no_dpki: true
```
ThetaSinner, line 3: I think we should make a decision about whether we're testing with this on or off. Maybe off makes sense right now because Holo are configuring their system that way for now. 

But equally, you could argue that this is the future direction and we should have at least one test that turns it on before Holo start using it.

Thoughts from everyone else?

ThetaSinner, line 3: Separately, this should also go into the "non CI" version of this file that gets used with Holoports. Currently, only the configured network services should be different I think

scenarios/validation_receipts/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    let action_hash: ActionHash = call_zome(
+        ctx,
+        "crud",
+        "create_sample_entry",
+        "this is a test entry value",
+        Some(Duration::from_secs(80)),
+    )?;
+
+    let response: Option<Record> = call_zome(
```
ThetaSinner, line 53: Is reading the entry back necessary for this test? I'm thinking probably not - because it should get read back faster than any peers can respond with signatures anyway so it's unlikely to make the timing line up nicely. May as well leave a few extra cycles for our node to do validation :)

scenarios/validation_receipts/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+        "get_sample_entry",
+        action_hash.clone(),
+        Some(Duration::from_secs(80)),
+    )?;
+
+    assert!(response.is_some(), "Expected record to be found");
+
+    let start = Instant::now();
+
+    'outer: loop {
```
ThetaSinner, line 65: Busy loop? I would say that it's acceptable to sleep for 10-25ms or something around there. It does affect the results slightly but I'd rather we correct for that in the future by recording timestamps on the validation receipts so we can get an accurate receive time without needing to query fast.

scenarios/validation_receipts/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+            Some(Duration::from_secs(80)),
+        )?;
+
+        // only check for complete if we actually get data back
+        // before any receipts are received, this is an empty vec
+        if !response.is_empty() {
+            let mut all_complete = true;
+
+            'inner: for set in response.iter() {
+                if !set.receipts_complete {
```
ThetaSinner, line 80: I wonder how hard we want to push the concurrency here. While we're waiting for receipts to come back, we pause adding load to the system. Another option would be to push out actions as fast as possible and each behavior would just see how many of its previous actions have now got their validation receipts and stop tracking those actions.

I think both scenarios are valid but the second is more similar to how an active agent will apply load. They won't wait for validation receipts, they'll just create actions as fast as it suits them and we want to know how well Holochain can keep up.

Maybe we should have both?

ThetaSinner (commented):
Looking good! I do have a few questions about this one but I'm open to any responses and I'd be happy that this adds great value as it is :)

scenarios/validation_receipts/src/main.rs
neonphog: Good call, I can remove this :+1:

scenarios/validation_receipts/src/main.rs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            if all_complete {
+                break 'outer;
+            }
+        }
+    }
+
+    reporter.add_custom(
+        ReportMetric::new("validation_receipts_complete_time")
```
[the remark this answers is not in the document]
neonphog: Actually, it's pretty low hanging fruit to just report the metrics separately, and that strictly gives us additional data.

conductor-config-ci.yaml
neonphog: I guess the simplest thing to do would be to not commit this change for now. Anyone can modify or use separate configs locally for whatever test scenario they want to use, and we can figure out what the default should be separately.

neonphog: done

neonphog: just removed this from the ci version for now

scenarios/validation_receipts/src/main.rs
neonphog: done

scenarios/validation_receipts/src/main.rs
neonphog: removed

scenarios/validation_receipts/src/main.rs
neonphog: done. Added example charts for both to the top-level description of this PR

scenarios/validation_receipts/src/main.rs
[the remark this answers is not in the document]
neonphog: done. And luckily the existing dashboard just picked up the extra tag and charted the extra lines without any changes.

ThetaSinner (approved):
Looks awesome!

PR #87, opened 2024-09-04
Nixos and direnv setup

cdunster:
This PR:
* Adds support for [direnv](https://direnv.net) and automatically loads the `devShell.default` from the nix flake.
* Fixes a small typo in the readme file.
* Allows scenarios to be ran locally on NixOS by fixing the openssl dependency.


flake.nix
```diff
@@ -88,6 +90,8 @@
             pkgs.darwin.apple_sdk.frameworks.CoreFoundation
           ];
 
+          OPENSSL_NO_VENDOR = 1;
```
ThetaSinner, line 93: I'm confused why this is needed. What should be happening is that libsodium (which is where I think the dependency is coming in, could be a couple of other places like sqlite) is set up to vendor openssl so that we get a version built for the current platform that is correctly configured for use with the Rust library. It shouldn't be looking for a system library.

I'm wondering if we're missing something else that is stopping that happening. E.g. this was required in Holonix -> https://github.com/holochain/holonix/blob/main/flake.nix#L139

We had a lot of trouble with platforms providing their own openssl when people built the software outside of Nix and so I'd really like to make sure that we're still vendoring like we intend to.

What do I need to do to reproduce the problem you ran into here? Is it entering the dev shell, or is it when you try to build Wind Tunnel project binaries from inside the shell?

ThetaSinner: If it's the latter, I'm up for setting up a derivation to build Wind Tunnel and run it on CI to make sure this keeps working. We're building parts of it that way but not all I think

ThetaSinner (approved):

PR #89, opened 2024-09-06
Improve countersigning scenario

ThetaSinner:
image


bindings/trycp_runner/src/common.rs
```diff
@@ -314,7 +314,7 @@ where
         let agent_name = agent_name.clone();
         async move {
             let logs = client
-                .download_logs(agent_name, Some(Duration::from_secs(180)))
+                .download_logs(agent_name, Some(Duration::from_secs(10 * 60)))
```
ThetaSinner, line 317: When the logs are outputting a lot of information it can take a few minutes to get everything

conductor-config-ci.yaml
```diff
@@ -1,3 +1,6 @@
+dpki:
+  dna_path: ~
+  no_dpki: true
```
ThetaSinner, line 3: - [ ] Needs @maackle's change to DPKI for test mode before this can be resolved

flake.nix
```diff
@@ -8,6 +8,7 @@
     holonix = {
       url = "github:holochain/holonix?ref=main";
       inputs = {
+        holochain.url = "github:holochain/holochain?ref=4159-countersigning-workflow-rework-debug";
```
ThetaSinner, line 11: - [ ] Needs a release of Holochain that has the countersigning fixes

scenarios/two_party_countersigning/src/main.rs
```diff
@@ -117,132 +119,109 @@ fn agent_behaviour_initiate(
...
-                            }
-                        }
+                    let session_times = my_preflight_response.request.session_times.clone();
+                    let session_timeout = Instant::now().add(Duration::from_millis(
+                        (session_times.end.as_millis() - session_times.start.as_millis()) as u64,
+                    ));
 
-                        // Add no new agents, that should only happen when we exhaust the list.
-                        Ok(Vec::with_capacity(0))
+                    match run_initiated_session(
```
ThetaSinner, line 192: Hideous diff, the code to the left and above has been moved to a function. The reason is to make sure that any errors that happen after the session has started won't cause the agent behavior to continue to the next iteration. We have to wait for the session to time out if something goes wrong!

scenarios/two_party_countersigning/src/main.rs
```diff
@@ -252,6 +231,102 @@ fn agent_behaviour_initiate(
...
+                    Signal::App { signal, .. } => signal,
+                    _ => {
+                        log::debug!("Received a signal that is not an app signal, listening for other signals.");
+                        continue;
+                    }
+                };
+
+                let other_response = match signal.clone().into_inner().decode::<Signals>() {
+                    Ok(Signals::Response(response))
+                        if session_fingerprint(response.request())?
```
ThetaSinner, line 265: This changed. Previously I was comparing the `app_entry_hash` which isn't unique and was causing crossed wires and so validation failures. This is how Holochain checks that sessions are equal and massively improves the scenario behavior.

scenarios/two_party_countersigning/src/main.rs
```diff
@@ -488,6 +563,15 @@ async fn complete_session(
     Ok(retry_count)
 }
 
+fn session_fingerprint(preflight_request: &PreflightRequest) -> anyhow::Result<Vec<u8>> {
+    let hash = holo_hash::encode::blake2b_256(
```
ThetaSinner, line 567: - [ ] Move to Holochain, it can be a function on the `PreflightRequest`

scenarios/two_party_countersigning/src/main.rs
```diff
@@ -117,132 +119,109 @@ fn agent_behaviour_initiate(
...
+                        Ok(_) => {
+                            // Completed successfully
+                        }
+                        Err(e) => {
+                            log::warn!(
+                                "Failed to run initiated session, waiting for it to time out: {:?}",
+                                e
+                            );
+
+                            // TODO this can be replaced with listening for the abandoned signal
```
ThetaSinner, line 215: - [ ] Timeout signals need to be handled in a couple of places through this test. Update this once Holochain is released with that signal defined

neonphog (dismissed):
lgtm : )

scenarios/two_party_countersigning/src/main.rs
```diff
@@ -507,10 +614,20 @@ async fn await_countersigning_success(
...
                     }
+                    Signal::System(SystemSignal::AbandonedCountersigning(eh))
+                        if eh == session_entry_hash =>
+                    {
+                        return Err(anyhow::anyhow!("Countersigning session was abandoned"));
+                    }
+                    Signal::System(SystemSignal::AbandonedCountersigning(_)) => {
+                        // This shouldn't happen because only one countersigning session can be active at a time. There's a bug if this log message shows up.
+                        log::warn!("Received an abandoned countersigning signal for a different session, listening for other signals.");
+                        continue;
```
cdunster, line 629: Given the comment then should we log this as an error instead?

scenarios/two_party_countersigning/Cargo.toml
```diff
@@ -12,6 +12,8 @@ log = { workspace = true }
 rmp-serde = { workspace = true }
 
 holochain_types = { workspace = true }
+holo_hash = { workspace = true }
+holochain_serialized_bytes = { workspace = true }
```
cdunster, line 16: Are these dependencies used? I see them added here but I don't see a `use` statement anywhere.

conductor-config-ci.yaml
```diff
@@ -1,6 +1,21 @@
+#danger_generate_throwaway_device_seed: true
+#dpki:
+#  dna_path: ~
+#  network_seed: wind-tunnel-test
+#  allow_throwaway_random_dpki_agent_key: true
+#  no_dpki: false
```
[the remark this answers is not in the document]
ThetaSinner: Ideally not removed no. I'd prefer that we can test either way. There's a conversation on #92 about ways to do that.

For now, enabling DPKI would require changing code to pass the config through to common helper functions so they can modify their behaviour around app installs. You can't provide an agent key if you're using DPKI. I'm also not sure that's entirely intentional behaviour around that, but that's another issue.

Basically, more work required yet to support testing with either config.

scenarios/two_party_countersigning/src/main.rs
ThetaSinner: Fair point, yes

scenarios/two_party_countersigning/src/main.rs
```diff
@@ -315,45 +390,48 @@ fn agent_behaviour_participate(
...
-                        ).await?;
-
-                        log::debug!("Accepted the incoming session, proceeding to commit.");
+                        match run_accepted_session(client, request, session_timeout, cell_id, app_port).await {
+                            Ok(retry_count) => {
+                                let elapsed = start.elapsed();
+
+                                log::debug!("Completed countersigning session with the initiating party.");
+
+                                let accepted_success = accepted_success.fetch_add(1, std::sync::atomic::Ordering::Acquire);
```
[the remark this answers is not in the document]
ThetaSinner: Yep, this is multi-threaded but each agent runs on a single thread and there's no actual parallelism within that agent's behaviour. Relaxed sounds sensible. I've updated all uses here

scenarios/two_party_countersigning/Cargo.toml
ThetaSinner: Nope, removed

conductor-config-ci.yaml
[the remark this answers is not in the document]
jost-s: > You can't provide an agent key if you're using DPKI. I'm also not sure that's entirely intentional behaviour around that, but that's another issue.

I agree, that's a confusing behavior that I've also noticed. It will need to be possible for Holo anyway and other people who generate their keys outside of Holochain app installation.

neonphog (approved):

PR #92, opened 2024-09-11
Holochain chc performance testing

c12i:


.github/workflows/test.yaml
```diff
@@ -245,4 +251,4 @@ jobs:
         with:
           name: scenario-bundles
           if-no-files-found: error
-          path: dist/*
+          path: dist/*
```
ThetaSinner, line 254: Please can newlines at the end of files be left?

conductor-config-ci.yaml
```diff
@@ -4,3 +4,4 @@ network:
     - type: webrtc
       signal_url: "ws://localhost:4423"
   bootstrap_service: "http://localhost:4422"
+# chc_url: "http://localhost:8181"
```
ThetaSinner, line 7: What are we going to do about needing different config for some scenarios?

.github/workflows/test.yaml
```diff
@@ -120,8 +120,14 @@ jobs:
 
       - name: Smoke test - write_read
```
ThetaSinner, line 121: This needs to be split so we run with and without CHC for this scenario

flake.nix
```diff
@@ -12,6 +12,8 @@
         flake-parts.follows = "flake-parts";
         crane.follows = "crane";
         rust-overlay.follows = "rust-overlay";
+        # TODO: Revert when new release of holochain is made with updated hc-sandbox cli
+        holochain.url = "github:holochain/holochain?ref=develop";
```
ThetaSinner, line 16: Is there a tag now that we could use here so we know when Holonix has caught up? I think 0.4.0-dev.24 should have what you needed

conductor-config-ci.yaml
c12i: I think the `embed_conductor_config` function might need to be refactored slightly, it should return a type rather than the yaml in the form of a string, so that values can be mutated by scenarios that need custom conductor config different from the one in the yaml files.

What do you think?

ThetaSinner: I'm not sure about the scenario modifying the config. We want to be able to run scenarios with different config. In this case that means we want to run with CHC on and off. We don't then want to be modifying the scenario code to run in different environments etc. 

What would be really nice is if you could provide a conductor configuration "overlay". So the `embed_conductor_config` would load a base configuration then at runtime you could configure which overlay to load. It might make sense to merge those as conductor configuration objects rather than yaml.

c12i: Sorry I am late, got distracted with other work

> I'm not sure about the scenario modifying the config. We want to be able to run scenarios with different config. In this case that means we want to run with CHC on and off. We don't then want to be modifying the scenario code to run in different environments etc.

I see what you mean, what would a good compromise be? perhaps having `-chc`variants of the existing yaml configs, not the cleanest approach.

Or rather than reading from the conductor config files, we programmatically build the conductor configs from the scenario, with an env var i.e `CHC_ENABLED` determining whether a `chc_url` should be set.

I am open to discuss this further

ThetaSinner: When I'm talking about overlays, I mean this kind concept -> https://carvel.dev/ytt/

Not necessarily that tool but something along those lines. So we could have a `with-chc.yaml` and a `with-dpki.yaml` that can choose to load. 

I think choosing to load that with an environment variable like `CHC_ENABLED` is good. So the current macro can load all the config files and embed them. Then we can choose at runtime which ones we merge with the base conductor config.

I know that YAML merging is a bit of a pain in general but if it's purely additive for now, then it shouldn't be so bad I hope.

What do you think?

c12i: I see, would the contents of the `with-chc.yaml` and `with-dpki.yaml` only contain the CHC and DPKI configurations?

ThetaSinner: Exactly that, they'd just bring in the piece that they configure. So we might do other things the same way, like different tunings of the network or the conductor itself.

We probably couldn't easily guarantee that a given combination of overlays makes sense but I'm okay with leaving that up to the user for now

c12i: Understood. The only question I have now is how to run scenarios with CHC enabled via trycp. I believe the holoports would need to have the reference implementation running?

ThetaSinner: I think we need to deploy somewhere that the HoloPorts can see yes. We have a couple of servers but nowhere ideal really. I think this might need to be a question we ask to techops

.github/workflows/test.yaml
```diff
@@ -74,20 +74,47 @@ jobs:
...
           nix develop .#ci -c bash -c "hc s clean && echo "1234" | hc s --piped create && echo "1234" | hc s --piped -f 8888 run &"
-          
+
           # Run the scenario for 5 seconds
           RUST_LOG=info nix run .#zome_call_single_value -- --connection-string ws://localhost:8888 --duration 5 --no-progress
-          
+
+          pkill hc && pkill holochain && pkill lair-keystore
+
+      - name: Smoke test - zome_call_single_value with CHC enabled
```
ThetaSinner, line 83: This scenario doesn't write data, is it intentional to have CHC enabled on it? I don't think it will touch CHC at all, so I guess that's a kind of baseline but I'd argue that we already have that from running one of the scenarios that writes data and doesn't have the CHC enabled

.github/workflows/test.yaml
```diff
@@ -74,20 +74,47 @@ jobs:
...
           nix develop .#ci -c bash -c "hc s clean && echo "1234" | hc s --piped create && echo "1234" | hc s --piped -f 8888 run &"
-          
+
           # Run the scenario for 5 seconds
           RUST_LOG=info nix run .#single_write_many_read -- --connection-string ws://localhost:8888 --duration 5 --no-progress
-          
+
+          pkill hc && pkill holochain && pkill lair-keystore
+
+      - name: Smoke test - single_write_many_read with CHC enabled
```
ThetaSinner, line 107: Similar with this one, it's only doing one write so it's not using the CHC much

.github/workflows/test.yaml
```diff
@@ -150,6 +219,19 @@ jobs:
 
           pkill hc && pkill holochain && pkill lair-keystore
 
+      - name: Smoke test - local_signals with CHC enabled
```
ThetaSinner, line 222: Also doesn't write data

.github/workflows/test.yaml
```diff
@@ -194,6 +309,26 @@ jobs:
           # Stop local network services
           pkill hc-run-local
 
+      - name: Smoke test - remote_call_rate with CHC enabled
```
ThetaSinner, line 312: Doesn't write data

bindings/trycp_runner/src/macros.rs
```diff
@@ -1,13 +1,34 @@
...
+                include_str!("../../../conductor-config/conductor-config-ci.yaml");
+
+            static CHC_CONDUCTOR_CONFIG: &str =
+                include_str!("../../../conductor-config/with_chc.yaml");
 
             match std::env::var("CONDUCTOR_CONFIG") {
-                Ok(value) if value == "CI" => CONDUCTOR_CONFIG_CI,
-                _ => CONDUCTOR_CONFIG,
+                Ok(value) if value == "CI" => {
+                    if std::env::var("CHC_ENABLED")
```
ThetaSinner, line 15: Would this be cleaner if it was a `let mut config = match` at the top level of this block and then the `if CHC_ENABLED` was a separate statement that might update the config before it is returned?

.github/workflows/test.yaml
```diff
@@ -177,10 +272,30 @@ jobs:
           # Stop local network services
           pkill hc-run-local
 
+      - name: Smoke test - trycp_write_validated with CHC enabled
+        run: |
+          set -x
+
+          # start the CHC reference implementation service
+          nix develop .#ci -c bash -c "hc-chc-service --port 8181 &"
```
ThetaSinner, line 280: Could a section be added to the README on running the CHC service? There are a couple already that just give a recommended command for things like `hc-run-local-services` or the sandbox. It would be useful to have these two steps documented where you start the service and ask the sandbox to use it

.github/workflows/test.yaml
c12i: Hmmm from my testing it did write to the CHC, but not as much removed

PR #100, opened 2024-09-25
Bump weekly and CI updates

ThetaSinner:


.github/workflows/performance.yaml
```diff
@@ -7,6 +7,10 @@ env:
   INFLUX_TOKEN: ${{ secrets.INFLUX_TOKEN }}
   WT_METRICS_DIR: "${{ github.workspace }}/telegraf/metrics"
 
+concurrency:
```
ThetaSinner, line 10: I think we want to prevent this running more than once at the same time?

flake.nix
```diff
@@ -100,8 +100,6 @@
         };
 
         devShells.ci = pkgs.mkShell {
-          inputsFrom = [ inputs'.holonix.devShells ];
```
ThetaSinner, line 103: Wrong syntax, doesn't do anything

cdunster (approved):

c12i (approved):

.github/workflows/performance.yaml
ThetaSinner: It's more about not being able to trust the results if mutliple runs are attempted concurrently on the machine

neonphog (approved):

ThetaSinner:
Closing in favour of #101 which has the same changes plus some more and a newer Holochain version

PR #103, opened 2024-09-30
feat: Results analysis

ThetaSinner:
I believe this is ready to go as a first draft. I do think that some of the way I'm expecting runIds to be consistent across nodes and that we can know information about how many agents participated and ended from one node is specific to Local/TryCP. This will need adjusting if we go to a distributed run approach. However, I believe that the data analysis is reasonable in a distributed case because that shouldn't be assuming anything about how the scenario was run.

Whether the metrics are exactly what we need is a question for when we start asking questions based on them. But we can get some basic insights and track performance over time with this.

This will still need integrating with the performance test pipeline. That's not straight forward enough that I think it should happen on this PR.


scenarios/two_party_countersigning/README.md
```diff
@@ -4,6 +4,20 @@
 
 This scenario tests the performance of `countersigning` operations.
 
+There are two roles, `initiate` and the `participate`. 
+
+The participants commit an entry to advertise that they are willing
+to participate in sessions. They listen for sessions and participate in one at a time. Three metrics are recorded:
```
cdunster, line 10:
```suggestion
to participate in sessions. They listen for sessions and participate one at a time. Three metrics are recorded:
```

summariser/README.md
```diff
@@ -0,0 +1,50 @@
...
+The summariser comes with some tooling for testing it. With a `run_summary.jsonl` that contains a run which you want to 
+use as a test case, you can run the following command:
+
+```shell
+RUST_LOG=debug cargo run --features test_data summariser
+```
+
+This will add:
+- The run summary to `summariser/test_data/1_run_summaries/<scenario-name>-<scenario run fingerprint>.json`
+- The raw that fetched from influx, as JSON to `summariser/test_data/2_query_results/<query fingerprint>.json`
```
cdunster, line 27: Raw data? or logs?

```suggestion
- The raw data fetched from influx, as JSON to `summariser/test_data/2_query_results/<query fingerprint>.json`
```

summariser/src/analyze.rs
```diff
@@ -0,0 +1,301 @@
...
+        .filter(col("time").is_not_null())
+        .sort(
+            ["time"],
+            SortMultipleOptions::default().with_maintain_order(true),
+        )
+        .group_by_dynamic(
+            col("time"),
+            [],
+            DynamicGroupOptions {
+                // every: Duration::parse(window_duration),
```
cdunster, line 76: Same as the line bellow

```suggestion
```

summariser/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+                    .await
+                    .context("Write validated summary")
+            }
+            .boxed(),
+        ),
+        "two_party_countersigning" => Some(
+            async move {
+                summarize_countersigning_two_party(client.clone(), summary.clone())
+                    .await
+                    .context("Countersigning, two party, report")
```
cdunster, line 85: Nitpick: all the others say "summary"

```suggestion
                    .context("Countersigning, two party, summary")
```

summariser/src/query.rs
```diff
@@ -0,0 +1,174 @@
...
+    summary: &RunSummary,
+) -> anyhow::Result<DataFrame> {
+    let q = ReadQuery::new(format!(
+        r#"SELECT value, zome_name, fn_name, agent FROM "windtunnel"."autogen"."wt.instruments.operation_duration" WHERE run_id = '{}' AND (operation_id = 'app_call_zome' OR operation_id = 'trycp_app_call_zome') AND is_error = 'false'"#,
+        summary.run_id
+    ));
+    log::debug!("Querying: {:?}", q);
+
+    #[cfg(feature = "query_test_data")]
+    if cfg!(feature = "query_test_data") {
```
cdunster, line 64: Is this line necessary after `#[cfg(...)]`? Or is it just to remove the unreachable code warning?

cdunster (changes requested):
Looking good, I think the main thing is about the `10e6` :slightly_smiling_face:

summariser/src/scenario/validation_receipts.rs
```diff
@@ -0,0 +1,51 @@
...
+            .context("Receipts complete timing")?,
+            receipts_complete_rate: partitioned_rate_stats(
+                receipts_complete.clone(),
+                "value",
+                "10s",
+                &["agent", "op_type"],
+            )
+            .context("Receipts complete rate")?,
+            error_count: zome_call_error_count(client, &summary).await?,
+        },
```
neonphog, line 49: I don't love how you have to manually write one of these for every scenario. I totally see you trying to make this as declarative as possible to remove friction/boilerplate, and we don't have to get all the way there on the first try... but some questions for how you see this evolving in the future:

- Do you think there's a way to specify this closer to the scenario? I.e. if part of the scenario trait was to define a handler function for how to generate the summary, so we'd not have to cross crates when writing scenarios?
- Is there a way to go all the way declarative, like by specifying whether it should be partitioned rate stats or timing stats somehow in the field itself and having the summariser parse the field names to know if and how it should summarize them?

summariser/README.md
```diff
@@ -0,0 +1,50 @@
...
+Start by running one or more scenarios with `--reporter influx-file` to create metrics on InfluxDB. You should find a
+`run_summary.jsonl` file created in the root of the project. The summariser reads this to find metrics.
+
+You can then run the summariser to generate a report:
+
+```shell
+RUST_LOG=info cargo run summariser 
+```
+
+This will create a new JSON file that summarises the results from the scenarios.
```
jost-s, line 14: Should this say `JSONL` everywhere where summaries are generated? It won't be valid JSON, I believe.

summariser/src/analyze.rs
```diff
@@ -0,0 +1,301 @@
+use crate::model::{
+    PartitionKey, PartitionRateStats, PartitionTimingStats, PartitionedRateStats,
+    PartitionedTimingStats, StandardRateStats, StandardTimingsStats, TimingTrend,
+};
+use anyhow::Context;
+use itertools::Itertools;
+use polars::frame::DataFrame;
+use polars::prelude::*;
+
+pub(crate) fn standard_timing_stats(
```
jost-s, line 10: These are legible enough to not require tests for them. For more complicated stats it would be good to have tests for the formulas.

jost-s (dismissed):
Yeah, whoo

framework/summary_model/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// the roles that were assigned across the network.
+    pub peer_end_count: usize,
+    /// The behaviour configuration
+    ///
+    /// This is the number of agents that were assigned to each behaviour.
+    ///
+    /// Note: This is only meaningful for single-conductor tests with the standard Wind Tunnel runner
+    /// or with the TryCP runner. In general, each node only sees the roles it was assigned and not
+    /// the roles that were assigned across the network.
+    pub behaviours: HashMap<String, usize>,
```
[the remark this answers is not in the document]
ThetaSinner: I'm not that keen on `_peer_count` because it's not just a count. I've renamed this to match what it's called in the runner `assigned_behaviours`. Better? If we were going to rename it to anything else here, I'd want to rename it in the runner too

framework/summary_model/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn set_peer_end_count(&mut self, peer_end_count: usize) {
+        self.peer_end_count = peer_end_count;
+    }
+
+    /// Add an environment variable
+    pub fn add_env(&mut self, key: String, value: String) {
+        self.env.insert(key, value);
+    }
+
+    pub fn fingerprint(&self) -> String {
```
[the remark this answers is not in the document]
ThetaSinner: Haha! Fair :) This was a function in a different crate and I moved it without thinking about docs. Fixed

scenarios/dht_sync_lag/src/main.rs
```diff
@@ -72,15 +75,17 @@ fn agent_behaviour_record_lag(
...
+        let lag_s = (metric
             .timestamp
             .clone()
             .duration_since(SystemTime::UNIX_EPOCH)
             .unwrap()
             .as_micros()
             - timed_entry.created_at.as_micros() as u128) as f64
-            / 1000.0;
-        let metric = metric.with_field("value", lag_ms);
+            / 10e6;
```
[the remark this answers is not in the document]
ThetaSinner: Yep, well spotted and so the test data for these will be invalid. I'll have to re-run them

scenarios/two_party_countersigning/README.md
ThetaSinner: I do mean "participate in" here, should this sentence be re-written to be clearer?

summariser/src/query.rs
ThetaSinner: The second question, exactly that. It's ugly but deals with the warnings. The clean way to do this would probably have been to add a proper CLI interface and not use compile flags. Should I do that?

ThetaSinner: Maybe I should do it on another PR though...

summariser/src/scenario/validation_receipts.rs
ThetaSinner: Yep, objections understood. I'm not that keen to have the analysis logic in the scenarios because they're meant to get uploaded/run on target nodes that aren't responsible for analyzing. So I'm picturing those being two separate steps and so it made sense to me to separate the code out. Because they're separate though, I think it would be valid to do one PR that adds a scenario and we can experiment with the scenario for a while and come back and add analysis when we're ready. The tool will just ignore scenarios it doesn't know how to create a summary for.

As for whether we can make this declarative? Maybe, I'm thinking that what I've done is the minimum and some scenarios are going to need to do more complicated things and examine a mixture of automatically captured "client" metrics along with custom metrics. I think we'll probably limit ourselves by going down that route, or end up with a complex DSL of some kind for asking what metrics we want. Or I'm wrong and this is as complicated as the analysis needs to get and maybe we'll see we can go that route in the near future. 

Basically, I had a reason for separating the code out from the scenarios but what we need next is probably going to depend on what tools we pick and how useful we find the data and whether we need more complex insights or get plenty of value from these few analysis functions... I really don't know :) 

Something I do have planned that would possible be more complicated if we put analysis information in the scenarios is creating a change summary between two reports. That would involve compiling all the scenarios together if the equivalent of these models lived in each scenario. Having them in this crate should make it easier to define a "difference" trait of some kind that can use these models. But then again, if it was all declarative maybe we wouldn't need custom models and could just use a field to type the various analysis outputs and run a standard diff over them... Maybe! :)

One way or another, I'm certainly open to this being evolved!

summariser/README.md
ThetaSinner: The final report is valid JSON, I'm serializing a `Vec` of reports so the file will contain an array at the top level. It's entirely possible I've mixed up my file extensions in a few places, but I don't think here

summariser/src/analyze.rs
ThetaSinner: Hmm, yeah we technically are testing them in the functional tests but you're right that it would be good to pass some data frames where the statistics it should calculate are easy to check against the actual results. That would be valid for these, they aren't that simple! :) I'm tempted to say though, let's see if this is any use before investing in unit tests for them on top of the functional tests

jost-s: Agreed 👍

jost-s (dismissed):

ThetaSinner:
Turns out there was a problem with the data data. I hadn't included the scenario name in the fingerprint so scenarios with the same settings were getting the same fingerprint... Which isn't right! So I've sorted that now and added some tests/checks to stop things like that happening again

summariser/src/scenario/validation_receipts.rs
neonphog: > come back and add analysis when we're ready

:+1:

> I think we'll probably limit ourselves by going down that route

I agree we'll probably want more specialized analysis for specific problems that come up. I almost wish I had never used splunk... It would be trivial to pick two or more run ids out of the data, then pick the first timestamp for each run id and calculate a field that is just time-since-start so we can overlay and manipulate the graphs however we want to get at the visualization or summary we want.

Unfortunately, what I've seen so far experimenting with openobserve, it is not that easy to manipulate.

---

Okay, cool. Things to keep thinking about as we move forward.

neonphog (approved):
Nice :+1:

scenarios/two_party_countersigning/README.md
cdunster: Oh, re-reading it now I do get that. I guess it means "participate in one session at a time"? That makes sense now :slightly_smiling_face:

framework/summary_model/src/lib.rs
```diff
@@ -97,21 +97,33 @@ impl RunSummary {
         self.env.insert(key, value);
     }
 
+    /// Computer a fingerprint for this run summary
```
cdunster, line 100: I assume this is a typo but I was wrong before :sweat_smile: 
```suggestion
    /// Compute a fingerprint for this run summary
```

framework/summary_model/src/lib.rs
```diff
@@ -97,21 +97,33 @@ impl RunSummary {
...
+    ///
+    /// The fingerprint is intended to uniquely identify the configuration used to run the scenario.
+    /// It uses the
+    ///     - Scenario name
+    ///     - Run duration
+    ///     - Assigned behaviours
+    ///     - Selected environment variables
+    ///     - Wind Tunnel version
+    ///
+    /// The fingerprint is computed using SHA3-256.
```
cdunster, line 110: I wonder if a link to the type makes sense here?

```suggestion
    /// The fingerprint is computed using [sha3::Sha3_256].
```

summariser/src/main.rs
```diff
@@ -10,6 +10,8 @@ pub(crate) mod filter;
 async fn main() -> anyhow::Result<()> {
     env_logger::init();
 
+    println!("{}", 1e6);
+
```
cdunster, line 14: I guess this is leftover from debugging?

```suggestion
```

cdunster (approved):
Some minor suggestion but LGTM.

summariser/src/main.rs
ThetaSinner: Haha, yes

cdunster (dismissed):

cdunster (approved):

PR #115, opened 2024-10-17
feat: add run_with_required_agents function for TryCP scenarios

cdunster:
The minimum number of required agents to run the scenario to completion. Can be overridden via an environment variable `MIN_REQUIRED_AGENTS`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -177,8 +177,8 @@ jobs:
           # Start a TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && start_trycp &"
 
-          # Run the scenario for 10 seconds
-          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#trycp_write_validated -- --targets targets-ci.yaml --instances-per-target 2 --duration 10 --no-progress
+          # Run the scenario for 30 seconds
+          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#trycp_write_validated -- --targets targets-ci.yaml --instances-per-target 2 --duration 30 --no-progress
 
           # Stop the TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && stop_trycp"
@@ -194,8 +194,8 @@ jobs:
           # Start a TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && start_trycp &"
 
-          # Run the scenario for 10 seconds
-          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#remote_call_rate -- --targets targets-ci.yaml --instances-per-target 2 --duration 10 --no-progress
+          # Run the scenario for 30 seconds
+          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#remote_call_rate -- --targets targets-ci.yaml --instances-per-target 2 --duration 30 --no-progress
 
           # Stop the TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && stop_trycp"
@@ -209,8 +209,8 @@ jobs:
           # Start a TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && start_trycp &"
 
-          # Run the scenario for 10 seconds
-          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#two_party_countersigning -- --targets targets-ci.yaml --behaviour initiate:1 --behaviour participate:1 --instances-per-target 2 --duration 10 --no-progress
+          # Run the scenario for 30 seconds
+          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#two_party_countersigning -- --targets targets-ci.yaml --behaviour initiate:1 --behaviour participate:1 --instances-per-target 2 --duration 30 --no-progress
 
           # Stop the TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && stop_trycp"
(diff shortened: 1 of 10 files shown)


ThetaSinner (approved):
Very nice, looks good

bindings/trycp_runner/src/common.rs
```diff
@@ -466,3 +467,27 @@ where
...
+pub fn run_with_required_agents<RV: UserValuesConstraint, V: UserValuesConstraint>(
+    definition: ScenarioDefinitionBuilder<RV, V>,
+    min_required_agents: usize,
+) -> anyhow::Result<()> {
+    let agents_at_completion = run(definition)?;
+
+    let min_required_agents = std::env::var("MIN_REQUIRED_AGENTS")
+        .ok()
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(min_required_agents);
```
cdunster, line 484: Do we care about any of these errors? Right now if `MIN_REQUIRED_AGENTS` is not set or is not a valid number then it just silently uses the passed in value. Is this desired behaviour?

.github/workflows/test.yaml
```diff
@@ -177,8 +177,8 @@ jobs:
           # Start a TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && start_trycp &"
 
-          # Run the scenario for 10 seconds
-          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#trycp_write_validated -- --targets targets-ci.yaml --instances-per-target 2 --duration 10 --no-progress
+          # Run the scenario for 30 seconds
+          RUST_LOG=warn CONDUCTOR_CONFIG="CI" MIN_PEERS=2 nix run .#trycp_write_validated -- --targets targets-ci.yaml --instances-per-target 2 --duration 30 --no-progress
```
neonphog, line 181: Is it these timings that have the test time go from [12m](https://github.com/holochain/wind-tunnel/actions/runs/11498921230/job/32005758024) to [43m](https://github.com/holochain/wind-tunnel/actions/runs/11501744059/job/32015030249?pr=115)?

And if so, are we okay with that?

neonphog (commented):
Other than the test time question, this looks good to me :+1:

.github/workflows/test.yaml
jost-s: I'm wondering what we've been running as tests. The smoke tests on a recent workflow run all output the error saying that the app couldn't be installed, like here https://github.com/holochain/wind-tunnel/actions/runs/11498921230/job/32005758024#step:24:87

Is that a false error message?

cdunster: @neonphog, I don't think so / hope not. There are only 5 tests that were changed from 10 to 30 seconds so I'd expect an increase of 100 seconds. I think that the large difference here is just because it was the first build and so there's no cache as even the Rust Clippy job went from ~2 minutes to 12.5 minutes.

cdunster: @jost-s, from what I saw running locally and with CI, 10 seconds is not enough time to even install the apps so I think that the error message is correct.

ThetaSinner: Sounds right to me, some of the scenarios need more time to get going, so fair enough if they need more time. I roughly checked using logs that I'd given about enough time but hadn't got this far yet. The scenario is absolutely meant to run at least one behavior loop. 

Maybe this would be better drive with a proper test harness at some point rather than the real runner, where we can guarantee a limited number of passes doesn't produce errors... but even that is tricky because the behavior is allowed to error :)

bindings/trycp_runner/src/common.rs
ThetaSinner: I'd probably just log a debug message in the unwrap to say that the default value is being used, then it's easy enough to figure out what's going on

.github/workflows/test.yaml
jost-s: If the app installation fails for a percentage of agents, should the test fail too? In the workflow run they're shown as successful.

ThetaSinner: The agent is meant to shut down if they run into a fatal error. So if the errors aren't fatal then they should get recorded.

That does still leave room for CI missing a breakage I know, but I don't want too much in the framework that is specific to CI if we can help it. The behavior functions need to be allowed to fail, because they validly do sometimes

ThetaSinner: Open to suggestions like this PR though, if we can make CI work better without adding too much to Wind Tunnel that might accidentally cause problems in real runs

jost-s: Is there another explanation of the vastly increased test run duration?

ThetaSinner: I don't think so, I believe the answer @cdunster gave is accurate. The CI relies quite a bit on caching, if you make changes that require everything to be rebuilt then it takes quite a bit longer. Changing a single scenario is fairly quick to build but changing the framework requires all scenarios to be rebuilt and takes longer

It might be that we can make the CI faster but I don't know by how much

neonphog (approved):

bindings/trycp_runner/src/common.rs
cdunster: I added a `debug` if it's not set at all because this is probably the expected path but then a `warn` if it's an invalid value.

bindings/trycp_runner/src/common.rs
```diff
@@ -466,3 +468,32 @@ where
...
+    min_required_agents: usize,
+) -> anyhow::Result<()> {
+    let agents_at_completion = run(definition)?;
+
+    let min_required_agents = std::env::var("MIN_REQUIRED_AGENTS")
+        .inspect_err(|_| debug!("MIN_REQUIRED_AGENTS not set, using default value"))
+        .ok()
+        .and_then(|v| {
+            v.parse()
+                .inspect_err(|_| warn!("Invalid MIN_REQUIRED_AGENTS value. Using default"))
```
ThetaSinner, line 487: Nice!

CHANGELOG.md
```diff
@@ -34,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   other words, the `agent` is not reported when calling a clone cell.
 - All metrics are now reported in seconds, as an `f64`. There were some types still using milliseconds which made reporting
   across scenarios more complex.
+- Increased TryCP test scenario duration to 30s in CI [Performance Workflow](.github/workflows/performance.yaml).
```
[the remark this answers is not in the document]
cdunster: Oops

PR #117, opened 2024-10-22
Fix: TryCP tests running on targets

cdunster:
This PR makes it so that the TryCP test scenarios in the `Performance` workflow are actually run on the targets defined in `targets.yaml` as they should be.

It also updates `trycp_client` and `trycp_api` versions to `0.17.0-dev.6` to use the `lair-keystore` in-process instead of the binary as there is an issue with getting the connection URL from the binary (see holochain/lair#137).

### TODO
- [x] Update changelog

### DIFF
--- a/.github/workflows/performance.yaml
+++ b/.github/workflows/performance.yaml
@@ -109,7 +109,7 @@ jobs:
           # Start a TryCP instance
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && start_trycp &"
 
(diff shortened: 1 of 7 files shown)


ThetaSinner (approved):

cdunster:
> Surprised we're seeing the arx issue without having upgraded that input. I just dealt with that on the binaries repo. It needs some tweaks to the derivation definition

Yeah, I updated the `tryorama` input so maybe it came from there.

ThetaSinner (approved):

PR #122, opened 2024-10-29
Make building consistent in nix

### LINKED ISSUE #97: SPIKE: Minimise differences between build script and Nix build
Currently it's possible to set up a project that will be accepted and built by the built script but not by Nix. This should be minimized to avoid confusing build errors when going from local development to CI

cdunster:
This PR is to close the spike #97.

It adds an assert during the building of scenarios to make sure that the scenario directory has the same name as the cargo package itself.

Other solutions were explored and can be seen in the comment in the spike: https://github.com/holochain/wind-tunnel/issues/97#issuecomment-2447146196

Closes #97

### DIFF
--- a/scenarios/scenario_build.rs
+++ b/scenarios/scenario_build.rs
@@ -1,9 +1,22 @@
 use happ_builder::{build_happs, required_tools_available, BuildOptions, HappBuilderResult};
+use std::{path::PathBuf, str::FromStr};
 
 /// Build the required DNA(s) and hApps for the scenario.
 /// The built DNA(s) will appear as `/dnas/<scenario-name>/<dna-name>.dna` from the project root.
 /// The built hApp(s) will appear as `/happs/<scenario-name>/<happ-name>.happ` from the project root.
 fn main() -> HappBuilderResult {
+    let scenario_dir = PathBuf::from_str(std::env!("CARGO_MANIFEST_DIR")).unwrap();
+    let dir_name = scenario_dir.file_name().unwrap().to_str().unwrap();
+    let package_name = std::env!("CARGO_PKG_NAME");
+    if package_name != dir_name {
+        // The directory name is used as the Nix package name which is in turn used with `cargo run`
+        // inside the Nix package. Therefore, panic during the build if the directory name does not
+        // match the package name.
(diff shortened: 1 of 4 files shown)


neonphog (approved):
I'm cool with this as an iterative improvement. We can discuss what else we might want to do in the future outside the scope of this.

neonphog (dismissed):

scenarios/scenario_build.rs
```diff
@@ -1,9 +1,24 @@
...
+/// The built hApp(s) will appear as `/happs/<scenario-name>/<happ-name>.happ` from the project root.
 fn main() -> HappBuilderResult {
+    let scenario_dir = PathBuf::from_str(std::env!("CARGO_MANIFEST_DIR")).unwrap();
+    let dir_name = scenario_dir.file_name().unwrap().to_str().unwrap();
+    let package_name = std::env!("CARGO_PKG_NAME");
+    if package_name != dir_name {
+        // The directory name is used as the nix package name which is in turn used with `cargo run`
+        // inside the nix package. Therefore, fail the build if the directory name does not match the
+        // package name.
+        anyhow::bail!(
```
ThetaSinner, line 17: Isn't a panic equally informative here? It's in a `main` so it's always at the top level with no context to track and it's a build script rather than a 'user-facing' command line tool. I'd prefer to avoid the `anyhow` dependency here

ThetaSinner (commented):
One preference comment, but the fix looks good to me, happy with that!

scenarios/scenario_build.rs
cdunster: Sure, I was torn because this way adds a dependency but it has a cleaner stderr:
```console
 Error: The package name of the scenario 'app_install' does not match the directory name 'app_installer'
```
vs a panic:
```console
thread 'main' panicked at scenarios/app_installer/../scenario_build.rs:17:9:
The package name of the scenario 'app_install' does not match the directory name 'app_installer'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ThetaSinner (approved):
:octocat: 🥳

PR #136, opened 2024-12-04
feat: add support for running scenarios on Nomad cluster

### LINKED ISSUE #132: Replace TryCP with Nomad for running scenarios
After the spike #111, the decision was made to replace TryCP with [Nomad](https://www.nomadproject.io). This issue is to track the progress of that replacement.

Nomad requires some ports to be open to communicate between the Nomad server and clients, see [here](https://developer.hashicorp.com/nomad/docs/install/production/requirements#ports-used).

We need to coordinate with Holo to make sure that the available Holo Ports are configured correctly to be used with Nomad.

AC:
* [ ] ~~Add the Nomad package to the NixOS definitions for the Wind Tunnel Holo Ports.~~
  * [ ] ~~This may require deploying a custom agent configuration on the Holo Port clients.~~
  * Edit: The Holo Ports will come later, the next step is: https://github.com/holochain/holochain/issues/4749
* [x] Package the scenarios as binaries that the Nomad Clients can reach. This might mean saving them as artefacts in a GitHub job and then using the Nomad `artifact` block to download them as part of a task.
* [x] Create a Nomad task to run the scenarios on the clients running on the Holo Hosts.
* [x] Get telemetry data and upload it to the InfluxDB instance for Wind Tunnel.
* [x] Support running the scenarios locally with Nomad.
* [x] Add documentation on how to run a scenario locally and on the Wind Tunnel Nomad cluster.

Follow-on work:
* #161
* #162

Out of scope:
* A new approach to packaging the scenarios. Stick with Nix for now.

cdunster:
This adds support for running Wind Tunnel scenarios using Nomad, both for running locally and on the dedicated Nomad cluster of test devices.

Closes #132

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -86,7 +86,7 @@ jobs:
           NOMAD_TOKEN: ${{ secrets.NOMAD_ACCESS_TOKEN }}
           NOMAD_VAR_scenario-url: ${{ steps.get-download-url.outputs.download-url }}
           NOMAD_VAR_run-id: ${{ github.run_id }}
-        run: |
+        run: |-
           nix run --impure --inputs-from . nixpkgs#nomad -- job run -var-file="nomad/var_files/${{ matrix.job-name || matrix.scenario-name }}.vars" nomad/run_scenario.nomad.hcl
 
           echo "Ran ${{ matrix.job-name }} with run ID ${{ github.run_id }}" >> $GITHUB_STEP_SUMMARY


nomad/run-app_install-local.nomad.hcl
```diff
@@ -0,0 +1,35 @@
...
+      }
+    }
+
+    task "run_scenario" {
+      driver = "raw_exec"
+      env {
+        RUST_LOG = "info"
+      }
+      config {
+        command = abspath("result/bin/app_install")
```
ThetaSinner, line 23: Will there be a way to have this reference a common path locally and remotely so that we can use the same scripts in dev as on a real environment?

nomad/run-app_install-local.nomad.hcl
```diff
@@ -0,0 +1,35 @@
...
+        args = [
+          "--connection-string", "ws://localhost:8888",
+          "--agents", "2",
+          "--behaviour", "minimal:1",
+          "--behaviour", "large:1",
+          "--duration", "5",
```
ThetaSinner, line 29: Should this default to a longer time when you're ready to merge this?

ThetaSinner (commented):
Tried this out, looking good!

nomad/run-app_install-local.nomad.hcl
cdunster: I'm hoping to allow this but I'm not yet sure what is the best way. I think at the very least it should be possible to provide either a path or a URL and if it's a URL then the binary gets downloaded as an artefact before being run.

nomad/run-app_install-local.nomad.hcl
cdunster: My idea (already in the latest version) is that if no duration is provided then it is not actually set and so it will use the default provided by the scenario. This could be an issue for neverending scenarios but I think it makes the most sense and we can always kill a running scenario at set the duration if we miss it by accident.

bindings/client/src/admin_websocket.rs
```diff
@@ -25,7 +26,11 @@ pub struct AdminWebsocketInstrumented {
 impl AdminWebsocketInstrumented {
     pub async fn connect(admin_url: impl ToSocketAddr, reporter: Arc<Reporter>) -> Result<Self> {
         let addr = admin_url.to_socket_addr()?;
-        Ok(AdminWebsocket::connect(addr)
+        let config = WebsocketConfig {
```
ThetaSinner, line 29: Did we require this globally or just on app install? I wonder if we should permit the client to override these timeouts per call so that we can be more specific. That's a problem for another time but I was just curious

ThetaSinner (commented):
This is looking great. I have a few minor comments and I'm also wondering if there is much/any value in updating the suggested commands in the README of each scenario that can be run this way. Do we want to provide the command that needs to be run when the scenario is used with Nomad as well as locally? Or is it enough for people to refer to the CI workflow and root readme?

nomad/run-app_install-local.nomad.hcl
cdunster: Done as part of 4ffd6688264b7198cabb017cf0a0c4b2c5bda218

.github/workflows/nomad.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+jobs:
+  run-scenarios:
+    name: Run ${{ matrix.job-name || matrix.scenario-name }}
+    runs-on: [self-hosted, wind-tunnel]
+    strategy:
+      fail-fast: false
+      # `job-name` defaults to `scenario-name` if not provided.
+      # `required-nodes` defaults to `1` if not provided.
+      matrix:
```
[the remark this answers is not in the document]
cdunster: Hhmmm, good point. I assumed that the wait for nodes step would solve this issue, but yeah, I think it would be possible that 2 runners pass the check at the same time and both start a job, which then causes the issue. I'll make the change.

.github/workflows/nomad.yaml
```diff
@@ -0,0 +1,91 @@
...
+            scenario-name: app_install
+
+          - scenario-name: dht_sync_lag
+            required-nodes: 2
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Bundle scenario
+        run: |
+          nix bundle --option sandbox false .#packages.x86_64-linux.${{ matrix.scenario-name }}
```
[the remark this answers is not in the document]
cdunster: I'm not actually sure, I saw that was [being used in the original workflow](https://github.com/holochain/wind-tunnel/blob/c0d44f850c8a95bc07a68841fc2e324e627145ff/.github/workflows/test.yaml#L280) to bundle them so I kept it but I also tried without and it worked so I wanted to look into it more and forgot

bindings/client/src/admin_websocket.rs
cdunster: I don't think this was actually the solution to the problem I was having so I will actually try and revert it and see

README.md
```diff
@@ -435,6 +435,166 @@ You can then start a second terminal and run one of the scenarios in the `scenar
...
+cp ./app_install-arx ./app_install
+```
+
+You now need to upload the scenario bundle to somewhere public so that the Nomad client can download it.
+This could be a GitHub release, a public file sharing services, or some other means, as long as it's publicly
+accessible.
+
+> [!Note]
+> Unlike when running locally in the section above, we cannot just pass a path because the path needs to be
+> accessible to the client and Nomad doesn't have native support for uploading artefacts.
```
mattyg, line 551:
```suggestion
> accessible to the client and Nomad doesn't have native support for uploading artifacts.
```
consistent with other uses of the word in this doc

mattyg (dismissed):
awesome

README.md
cdunster: Yeah, I wasn't sure about this. British English is mainly used in this repo, so I stuck with it, but when referencing other tools like GitHub and Nomad, they use US English, so I stuck with that :confounded:

.github/workflows/nomad.yaml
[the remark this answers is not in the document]
cdunster: Done: f3353d09af569b5dc9007a2f514725fd4d781e04

.github/workflows/nomad.yaml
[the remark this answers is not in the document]
cdunster: Oh, the comment in the original workflow says:

> Disable sandbox because hApp builds aren't done by Nix and need network access.

I'm not sure if this still stands as it seemed to work when I tried it would disabling the sandbox.

README.md
```diff
@@ -435,6 +435,166 @@ You can then start a second terminal and run one of the scenarios in the `scenar
...
+
+To run a Wind Tunnel scenario on the Nomad cluster from the command-line, first enter the Nix `devShell`
+with `nix develop` or [install Nomad](https://developer.hashicorp.com/nomad/install) locally.
+
+You also need to set the `NOMAD_ADDR` environment variable to `https://nomad-server-01.holochain.org:4646`
+and `NOMAD_CACERT` to `./nomad/server-ca-cert.pem`, which are both set by the Nix `devShell`.
+
+The final environment variable that needs to be set and is **not** set by the `devShell` is `NOMAD_TOKEN`
+which needs to be set to a token with the correct permissions, for now it is fine to just use the admin
+token found in BitWarden under `Nomad Server Bootstrap Token`.
```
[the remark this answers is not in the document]
cdunster: Done: a22183ed05a3a48b97611bd896c2805847fc568b

bindings/client/src/admin_websocket.rs
cdunster: Reverted in: 50307249247befc300bd640bced1c77aa44114c1. Let's see if it still works.

cdunster: It still seems to work with this change reverted.

README.md
ThetaSinner: No strong preference. I know what it means either way :)

PR #147, opened 2025-02-28
feat: add kitsune binding & sample scenario

### LINKED ISSUE #137: Write host implementation for a thin messaging app built on top of Kitsune2
Write an in-memory host implementation around kitsune2.

Wind Tunnel would like to run this as a library so it can be embedded in scenario binaries.

Simplest op structure, for example text messages encoded as bytes. Everything is constructed and kept in memory during the test.

AC:
- [ ] Op store needs to be wrapped to inform the space that ops have been processed.
- [ ] Assert ops have come in within a specific time window, based on `created_at` and `stored_at` comparison.
- [ ] No lair dependency. Agent private key can be generated and stored in plain text on disk.
- [ ] Single space, hashed off of user input or the wind tunnel run-id.
- [ ] Thin wrapper (`src/bin`) for testing during development or showcasing. All the functionality is a rust lib.
- [ ] Can bootstrap URL be passed into WT scenario through runner properties?
- [ ] Create binding in WindTunnel that exposes a handle to an in-memory instance of the host app.
- [ ] Write a WindTunnel runner for running this host implementation

jost-s:
This looks like a huge PR, ~~but about half of it is the op store implementation, and that is largely doing the same as the memory op store. I haven't just copied that over, because I wanted to understand what I was doing, but it turned out nearly identical.~~, and it no longer has the replicated op store implementation, so it's just huge. Sorry! I'm stopping here and keep the rest for the scenario issue.

I need to update the bootstrap server to run on a set port first to enable the test on CI.

resolves #137

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -26,10 +26,10 @@ jobs:
         if: runner.os == 'Linux'
         uses: AdityaGarg8/remove-unwanted-software@v2
         with:
-          remove-dotnet: 'true'
-          remove-android: 'true'
-          remove-codeql: 'true'
-          remove-docker-images: 'true'
+          remove-dotnet: "true"
+          remove-android: "true"
+          remove-codeql: "true"
+          remove-docker-images: "true"
 
       - name: Install nix
         uses: cachix/install-nix-action@v30
@@ -69,14 +69,18 @@ jobs:
         run: |
           nix develop .#ci -c bash -c "source ./scripts/checks.sh && check_rust_fmt"
 
-      - name: Build and unit tests
-        run: |
-          nix build .#workspace
-
       - name: Lint Rust
         run: |
           # Currently the only check is clippy, could bundle the other checks into the flake and remove the steps above?
           nix flake check --all-systems
+
+      - name: Run unit tests
+        run: |
+          nix develop .#ci -c bash -c "cargo test --workspace --all-targets"
+
+      - name: Build
+        run: |
+          nix build .#workspace
 
       - name: Smoke test - zome_call_single_value
         run: |
@@ -250,6 +254,19 @@ jobs:
           nix develop .#ci -c bash -c "source ./scripts/trycp.sh && stop_trycp"
           # Stop local network services
           pkill hc-run-local
+
+      - name: Smoke test - kitsune continuous flow
+        run: |
+          set -x
+
+          # Start local bootstrap and signal server
+          nix develop .#ci -c bash -c "kitsune2-bootstrap-srv --listen 127.0.0.1:30000 &"
+
+          # Run the scenario
+          RUST_LOG=warn cargo run -p kitsune_continuous_flow -- --bootstrap-server-url http://127.0.0.1:30000 --signal-server-url ws://127.0.0.1:30000 --duration 15 --agents 2
+
+          # Stop servers
+          pkill kitsune2-bootst
 
       - name: Build scenario bundles
         if: runner.os == 'Linux'
--- a/bindings/kitsune_client/Cargo.toml
+++ b/bindings/kitsune_client/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "kitsune_client_instrumented"
+version = "0.4.0-alpha.1"
+edition = "2021"
+
+[dependencies]
+anyhow = { workspace = true }
+bytes = { workspace = true }
+log = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+sha3 = { workspace = true }
+tokio = { workspace = true }
+
+kitsune2 = { workspace = true }
+kitsune2_api = { workspace = true }
+kitsune2_core = { workspace = true }
+kitsune2_gossip = { workspace = true }
+kitsune2_transport_tx5 = { workspace = true }
+
+wind_tunnel_core = { workspace = true }
+wind_tunnel_instruments = { workspace = true }
+wind_tunnel_instruments_derive = { workspace = true }
+
+[dev-dependencies]
+env_logger = { workspace = true }
+kitsune2_bootstrap_srv = "0.0.1-alpha6"
+
+[lints]
+workspace = true
--- a/bindings/kitsune_client/src/lib.rs
+++ b/bindings/kitsune_client/src/lib.rs
@@ -1,0 +1,324 @@
+use anyhow::Context;
+use bytes::Bytes;
+use kitsune2::default_builder;
+use kitsune2_api::{
+    BoxFut, Builder, DhtArc, DynKitsune, DynLocalAgent, DynSpace, DynSpaceHandler, K2Result,
+    KitsuneHandler, LocalAgent, OpId, SpaceHandler, SpaceId, StoredOp, Timestamp,
+};
+use kitsune2_core::{
+    factories::config::{CoreBootstrapConfig, CoreBootstrapModConfig},
+    Ed25519LocalAgent,
+};
+use kitsune2_gossip::{K2GossipConfig, K2GossipModConfig};
+use kitsune2_transport_tx5::config::{Tx5TransportConfig, Tx5TransportModConfig};
+use op_store::{DynWtOpStore, WtOp, WtOpStore, WtOpStoreFactory};
+use std::{sync::Arc, time::Duration};
+use tokio::sync::Mutex;
+use wind_tunnel_instruments::prelude::{ReportMetric, Reporter};
+use wind_tunnel_instruments_derive::wind_tunnel_instrument;
+
+mod op_store;
+
+#[derive(Debug)]
+struct WtSpaceHandler;
+impl SpaceHandler for WtSpaceHandler {}
+
+#[derive(Debug)]
+struct WtKitsuneHandler;
+impl KitsuneHandler for WtKitsuneHandler {
+    fn create_space(&self, _space: SpaceId) -> BoxFut<'_, K2Result<DynSpaceHandler>> {
+        let space_handler: DynSpaceHandler = Arc::new(WtSpaceHandler);
+        Box::pin(async move { Ok(space_handler) })
+    }
+}
+
+#[derive(Debug)]
+struct State {
+    agent: DynLocalAgent,
+    op_store: DynWtOpStore,
+    space: DynSpace,
+    _kitsune: DynKitsune,
+}
+
+/// A Kitsune2 app for running performance tests in WindTunnel.
+#[derive(Debug)]
+pub struct WtChatter {
+    state: Arc<Mutex<State>>,
+    reporter: Arc<Reporter>,
+}
+
+impl WtChatter {
+    /// Construct an instance.
+    pub async fn create(
+        bootstrap_server_url: &str,
+        signal_server_url: &str,
+        space_id: &str,
+        reporter: Arc<Reporter>,
+    ) -> anyhow::Result<Self> {
+        let agent = Arc::new(Ed25519LocalAgent::default());
+        agent.set_tgt_storage_arc_hint(DhtArc::FULL);
+        // Counter to common practice, an op store has to be created first and passed
+        // to the factory constructor, to keep a handle to the typed WtOpStore in the chatter
+        // instance. This store instance is also used to instantiate the dummy factory, which
+        // will simply return the same store at the time of calling it during space creation.
+        let op_store = Arc::new(WtOpStore::new(agent.agent().clone(), reporter.clone()));
+        let kitsune_builder = Builder {
+            op_store: Arc::new(WtOpStoreFactory::new(op_store.clone())),
+            ..default_builder()
+        }
+        .with_default_config()?;
+        kitsune_builder
+            .config
+            .set_module_config(&CoreBootstrapModConfig {
+                core_bootstrap: CoreBootstrapConfig {
+                    server_url: bootstrap_server_url.to_string(),
+                    ..Default::default()
+                },
+            })?;
+        kitsune_builder
+            .config
+            .set_module_config(&Tx5TransportModConfig {
+                tx5_transport: Tx5TransportConfig {
+                    server_url: signal_server_url.to_string(),
+                    signal_allow_plain_text: true,
+                    ..Default::default()
+                },
+            })?;
+        kitsune_builder
+            .config
+            .set_module_config(&K2GossipModConfig {
+                k2_gossip: K2GossipConfig {
+                    initiate_interval_ms: 1000,
+                    min_initiate_interval_ms: 900,
+                    ..Default::default()
+                },
+            })?;
+        let kitsune = kitsune_builder.build().await?;
+        kitsune.register_handler(Arc::new(WtKitsuneHandler)).await?;
+        // This will call the op store factory's `create` method.
+        let space = kitsune
+            .space(SpaceId::from(Bytes::copy_from_slice(space_id.as_bytes())))
+            .await?;
+
+        log::info!("created chatter with id {}", agent.agent());
+
+        let state = Arc::new(Mutex::new(State {
+            agent,
+            op_store,
+            space,
+            _kitsune: kitsune,
+        }));
+
+        Ok(Self { state, reporter })
+    }
+
+    /// Join the WindTunnel space.
+    #[wind_tunnel_instrument]
+    pub async fn join_space(&self) -> anyhow::Result<()> {
+        let state_lock = self.state.lock().await;
+        state_lock
+            .space
+            .local_agent_join(state_lock.agent.clone())
+            .await?;
+
+        // Wait for agent to publish their info to the bootstrap & peer store.
+        tokio::time::timeout(Duration::from_secs(20), async {
+            loop {
+                tokio::time::sleep(Duration::from_millis(1000)).await;
+                let maybe_agent = match state_lock
+                    .space
+                    .peer_store()
+                    .get(state_lock.agent.agent().clone())
+                    .await
+                {
+                    Ok(maybe_agent) => maybe_agent,
+                    Err(err) => {
+                        log::error!("failure to query peer store: {err}");
+                        continue;
+                    }
+                };
+                if maybe_agent.is_some() {
+                    break;
+                }
+            }
+        })
+        .await
+        .context("failure to join space")
+    }
+
+    /// Say a message, so that it will be gossiped to all peers.
+    #[wind_tunnel_instrument]
+    pub async fn say(&self, message: &str) -> anyhow::Result<Vec<OpId>> {
+        let state = self.state.lock().await;
+        let message_op = WtOp::new(Timestamp::now(), message.into());
+        let message_ids = state
+            .op_store
+            .store_op(message_op)
+            .await
+            .context("failure to write ops to the store")?;
+        state
+            .space
+            .inform_ops_stored(
+                message_ids
+                    .clone()
+                    .into_iter()
+                    .map(|message_id| StoredOp {
+                        created_at: Timestamp::now(),
+                        op_id: message_id,
+                    })
+                    .collect(),
+            )
+            .await?;
+        log::info!("agent {} said {}", state.agent.agent(), message);
+
+        self.reporter.add_custom(
+            ReportMetric::new("message said")
+                .with_tag("agent_id", state.agent.agent().to_string())
+                .with_field("message_said", message_ids.len() as u32),
+        );
+
+        Ok(message_ids)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use kitsune2_bootstrap_srv::{BootstrapSrv, Config};
+    use std::time::{Duration, Instant};
+    use wind_tunnel_core::prelude::ShutdownHandle;
+    use wind_tunnel_instruments::{ReportConfig, Reporter};
+
+    pub(crate) fn test_reporter() -> Arc<Reporter> {
+        let runtime = tokio::runtime::Handle::current();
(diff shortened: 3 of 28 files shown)


.github/workflows/test.yaml
```diff
@@ -251,6 +251,22 @@ jobs:
           # Stop local network services
           pkill hc-run-local
 
+      - name: Smoke test - kitsune
+        run: |
+          set -x
+
+          # Start local bootstrap server
+          nix develop .#ci -c bash -c "kitsune2-bootstrap-srv &"
```
jost-s, line 259: The bootstrap needs to be updated to accept a port for this to work.

bindings/kitsune_client/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+edition = "2021"
+
+[dependencies]
+anyhow = { workspace = true }
+bytes = "1.10.0"
+log = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+sha3 = { workspace = true }
+tokio = { workspace = true }
```
jost-s, line 13: This always includes full features. Maybe that's fine for WT.

bindings/kitsune_runner/src/cli.rs
```diff
@@ -0,0 +1,73 @@
...
+    #[clap(long)]
+    pub bootstrap_server_url: String,
+
+    /// The signal server URL.
+    #[clap(long)]
+    pub signal_server_url: String,
+
+    /// The number of agents to run.
+    #[clap(long)]
+    pub agents: Option<usize>,
```
jost-s, line 19: This is a debatable option. If tests are running on the same machine, it makes sense. When tests are run on multiple machines, not so much.

bindings/kitsune_runner/src/common.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::{KitsuneAgentContext, KitsuneRunnerContext};
+use anyhow::Context;
+use kitsune_client_instrumented::WtChatter;
+use serde::{Deserialize, Serialize};
+use wind_tunnel_runner::prelude::{
+    AgentContext, HookResult, ScenarioDefinitionBuilder, WindTunnelResult,
+};
+
+#[derive(Debug, Deserialize, Serialize)]
+struct KitsuneServerUrls {
```
jost-s, line 12: Bootstrap and signal server URL are passed as CLI arguments and then combined into the connection string.

kitsune_scenario/src/main.rs
jost-s, line 1: Very rudimentary scenario which will be fleshed out in the next issue.

.github/workflows/test.yaml
```diff
@@ -251,6 +251,22 @@ jobs:
           # Stop local network services
           pkill hc-run-local
 
+      # - name: Smoke test - kitsune
+      #   run: |
+      #     set -x
+
+      #     # Start local bootstrap server
+      #     nix develop .#ci -c bash -c "kitsune2-bootstrap-srv &"
```
jost-s, line 259: If multiple agents should be tested with a local bootstrap like this, the bootstrap needs to be updated to accept a port.

bindings/kitsune_client/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+        // let signal_server = tokio::task::spawn(
+        //     SbdServer::new(Arc::new(sbd_server::Config {
+        //         bind: vec!["127.0.0.1:0".to_string()],
+        //         ..Default::default()
+        //     })), // .await
+        // )
+        // .await
+        // .unwrap()
+        // .unwrap();
+        let signal_server_url = format!("ws://localhost:{}", signal_server.bind_addrs()[0].port());
```
steveej, line 220: could `localhost` potentially resolve the IPv6 variant `::1`?

```suggestion
        let signal_server_url = format!("ws://127.0.0.1:{}", signal_server.bind_addrs()[0].port());
```

bindings/kitsune_client/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+                    ..Default::default()
+                },
+            })?;
+        kitsune_builder
+            .config
+            .set_module_config(&K2GossipModConfig {
+                k2_gossip: K2GossipConfig {
+                    initiate_interval_ms: 1000,
+                    min_initiate_interval_ms: 900,
+                    // round_timeout_ms: 1000,
```
mattyg, line 93:
```suggestion
```

bindings/kitsune_client/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        .unwrap();
+        // let signal_server = tokio::task::spawn(
+        //     SbdServer::new(Arc::new(sbd_server::Config {
+        //         bind: vec!["127.0.0.1:0".to_string()],
+        //         ..Default::default()
+        //     })), // .await
+        // )
+        // .await
+        // .unwrap()
+        // .unwrap();
```
mattyg, line 220: The commented code in this function still needed?

bindings/kitsune_client/src/op_store.rs
```diff
@@ -0,0 +1,381 @@
...
+        let excluded_op_1 = MemoryOp {
+            created_at: Timestamp::from_micros(0),
+            op_data: vec![1],
+        };
+        let excluded_op_2 = MemoryOp {
+            created_at: Timestamp::from_micros(0),
+            op_data: vec![255],
+        };
+
+        // Store op to be exluded first
```
mattyg, line 294:
```suggestion
        // Store op to be excluded first
```

jost-s:
Thanks for reviewing, but this is still a draft.

bindings/kitsune_runner/src/common.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/// Parse cli argument "connection-string" for bootstrap and signal server URLs and return
+/// them separately.
+pub fn get_server_urls(
+    ctx: &AgentContext<KitsuneRunnerContext, KitsuneAgentContext>,
+) -> anyhow::Result<(String, String)> {
+    let connections =
+        serde_json::from_str::<KitsuneServerUrls>(ctx.runner_context().get_connection_string())
+            .context("failed to parse connectino string into bootstrap and server URL")?;
```
mattyg, line 24:
```suggestion
            .context("failed to parse connection string into bootstrap and server URL")?;
```

.github/workflows/test.yaml
```diff
@@ -251,6 +251,19 @@ jobs:
...
+          set -x
+
+          # Start local bootstrap and signal server
+          nix develop .#ci -c bash -c "kitsune2-bootstrap-srv --listen 127.0.0.1:30000 &"
+
+          # Run the scenario
+          RUST_LOG=warn cargo run -p kitsune_scenario -- --bootstrap-server-url http://127.0.0.1:30000 --signal-server-url ws://127.0.0.1:30000 --duration 40 --agents 2
+
+          # Stop servers
+          pkill kitsune2-bootst
```
mattyg, line 265:
```suggestion
          pkill kitsune2-bootstrap-srv
```

jost-s:
Yes, my bad. I appreciate your review, and several of your findings are valid despite this being a draft.

.github/workflows/test.yaml
jost-s: I did this because it's what pkill completes it to when hitting <TAB>, and the `hc-run-local-services` higher up in the file is abbreviated too.

.github/workflows/test.yaml
```diff
@@ -69,15 +69,19 @@ jobs:
         run: |
           nix develop .#ci -c bash -c "source ./scripts/checks.sh && check_rust_fmt"
 
-      - name: Build and unit tests
```
jost-s, line 72: Switched this with clippy step

.github/workflows/test.yaml
```diff
@@ -69,15 +69,19 @@ jobs:
...
-          nix build .#workspace
-
       - name: Lint Rust
         run: |
           # Currently the only check is clippy, could bundle the other checks into the flake and remove the steps above?
           nix flake check --all-systems
 
+      - name: Run unit tests
+        run: |
+          nix develop .#ci -c bash -c "cargo test --workspace --all-targets"
```
jost-s, line 79: New step to run all Rust tests

bindings/kitsune_client/src/op_store.rs
```diff
@@ -0,0 +1,379 @@
...
+    agent_id: AgentId,
+}
+
+impl OpStore for WtOpStore {
+    fn process_incoming_ops(&self, op_list: Vec<Bytes>) -> BoxFut<'_, K2Result<Vec<OpId>>> {
+        Box::pin(async {
+            let inserted_op_ids = self.inner.process_incoming_ops(op_list).await?;
+            let amount = inserted_op_ids.len();
+            log::info!("{} ops have come in to {}", amount, self.agent_id);
+            if amount > 0 {
```
jost-s, line 65: `process_incoming_ops` is frequently called with 0 ops. I think that can be improved in kitsune's gossip protocol. For the time being only register an operation when ops have actually come in.

flake.nix
```diff
@@ -107,16 +116,27 @@
         };
 
         devShells.ci = pkgs.mkShell {
+          nativeBuildInputs = [
+            pkgs.openssl
+            pkgs.rustPlatform.bindgenHook
+            pkgs.zlib
+          ];
           packages = [
+            pkgs.cmake
```
jost-s, line 125: I don't exactly know if all of these pkgs are actually required. I've rebuilt a couple of times locally to narrow them down, but had some "non-determinism" gremlins in my machine and it became too time-consuming to figure this out.

framework/instruments/src/lib.rs
```diff
@@ -53,7 +52,7 @@ impl ReportConfig {
 
     pub fn init_reporter(
         self,
-        runtime: &Runtime,
+        runtime: &tokio::runtime::Handle,
```
jost-s, line 55: I ended up having to change this to accept a Handle instead of a full Runtime, because the `wind_tunnel_instrument` macro expects an `Arc<Reporter>` field instead of an `Option<Arc<Reporter>>`. `Option::None` was used before in unit tests to circumvent this issue, but that collided with the macro.

happ_builder/src/lib.rs
```diff
@@ -405,9 +405,7 @@ fn print_rerun_for_package(package_dir: &Path) {
     walkdir::WalkDir::new(package_dir.join("src"))
         .into_iter()
         .filter_map(|e| e.ok())
-        .filter(|e| {
-            e.file_type().is_file() && e.path().extension().map_or(false, |ext| ext == "rs")
-        })
+        .filter(|e| e.file_type().is_file() && e.path().extension().is_some_and(|ext| ext == "rs"))
```
jost-s, line 408: This was a clippy fix

scenarios/kitsune_continuous_flow/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "kitsune_continuous_flow"
```
jost-s, line 2: This'll be scenario where messages are created in bursts of x messages every y seconds. Anyone offers a better name?

scenarios/kitsune_continuous_flow/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+        Ok(())
+    })
+}
+
+fn behavior(
+    ctx: &mut AgentContext<KitsuneRunnerContext, KitsuneAgentContext>,
+) -> anyhow::Result<()> {
+    let message = format!(
+        "message_{}",
+        std::time::UNIX_EPOCH
```
ThetaSinner, line 18: With a value like this, it's possible multiple agents might produce an op with the same id? Should the agent id be included here?

scenarios/kitsune_continuous_flow/src/main.rs
```diff
@@ -0,0 +1,39 @@
...
+            .elapsed()
+            .expect("time went backwards")
+            .as_millis()
+    );
+    say(ctx, &message)?;
+    ctx.runner_context().executor().execute_in_place(async {
+        tokio::time::sleep(Duration::from_secs(20)).await;
```
ThetaSinner, line 25: Maybe make this a configuration option for the test through an env var?

ThetaSinner: Although, we're doing that stuff on a later ticket right?

ThetaSinner (commented):
Looking good! I've left quite a few comments but I think that's a symptom of the size of the PR and it's actually mostly in good shape :)

scenarios/kitsune_continuous_flow/src/main.rs
jost-s: Yes, exactly. I haven't spent a lot of time on refining the scenario and its parameters etc., because that's part of the next issue.

scenarios/kitsune_continuous_flow/Cargo.toml
jost-s: Yeah, I'm thinking there'll be more kitsune scenarios and `op_sync` will apply to all of them in a way.

.github/workflows/test.yaml
```diff
@@ -251,6 +255,19 @@ jobs:
...
 
+      - name: Smoke test - kitsune
+        run: |
+          set -x
+
+          # Start local bootstrap and signal server
+          nix develop .#ci -c bash -c "kitsune2-bootstrap-srv --listen 127.0.0.1:30000 &"
+
+          # Run the scenario
+          RUST_LOG=warn cargo run -p kitsune_scenario -- --bootstrap-server-url http://127.0.0.1:30000 --signal-server-url ws://127.0.0.1:30000 --duration 40 --agents 2
```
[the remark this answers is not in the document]
jost-s: Yes, it's in seconds. I can bring it down for sure, there's quite a wait in the actual scenario and I haven't spent much time working on the scenario itself.

bindings/kitsune_client/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    #[wind_tunnel_instrument]
+    pub async fn say(&self, message: &str) -> anyhow::Result<Vec<OpId>> {
+        let state = self.state.lock().await;
+        let message_op = MemoryOp::new(Timestamp::now(), message.into());
+        let message_ids = state
+            .space
+            .op_store()
+            .process_incoming_ops(vec![message_op.into()])
```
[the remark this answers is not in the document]
jost-s: Absolutely. I've been thinking that too, dunno why I haven't done it.

bindings/kitsune_client/src/op_store.rs
```diff
@@ -0,0 +1,379 @@
...
+        &self,
+        _builder: Arc<Builder>,
+        space: SpaceId,
+    ) -> BoxFut<'static, K2Result<DynOpStore>> {
+        let agent_id = self.agent_id.clone();
+        let reporter = self.reporter.clone();
+        Box::pin(async move {
+            // Create a memory op store. The passed in builder cannot be used for this, because
+            // it would call this create function recursively ad infinitum.
+            let builder = Arc::new(default_builder().with_default_config()?);
```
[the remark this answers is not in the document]
jost-s: `MemOpStoreFactory::create().create(...)` requires a builder to be passed in. It comes out the same.

bindings/kitsune_client/src/op_store.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+impl OpStore for WtOpStore {
+    fn process_incoming_ops(&self, op_list: Vec<Bytes>) -> BoxFut<'_, K2Result<Vec<OpId>>> {
+        Box::pin(async {
+            let inserted_op_ids = self.inner.process_incoming_ops(op_list).await?;
+            let amount = inserted_op_ids.len();
+            log::info!("{} ops have come in to {}", amount, self.agent_id);
+            if amount > 0 {
+                let operation_record = OperationRecord::new("incoming_message".to_string());
```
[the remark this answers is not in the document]
jost-s: Only using the custom metric does not register the operation when reporting with the in memory reporter. That's why I had this here. It's kinda funny I agree.

bindings/kitsune_client/src/op_store.rs
```diff
@@ -0,0 +1,379 @@
...
+    reporter: Arc<Reporter>,
+    agent_id: AgentId,
+}
+
+impl OpStore for WtOpStore {
+    fn process_incoming_ops(&self, op_list: Vec<Bytes>) -> BoxFut<'_, K2Result<Vec<OpId>>> {
+        Box::pin(async {
+            let inserted_op_ids = self.inner.process_incoming_ops(op_list).await?;
+            let amount = inserted_op_ids.len();
+            log::info!("{} ops have come in to {}", amount, self.agent_id);
```
jost-s, line 64: You can run with info level traces to see that there are 0 incoming ops.

scenarios/kitsune_continuous_flow/src/main.rs
jost-s: Yes! Agent index should work too and is easier to read.

bindings/kitsune_client/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    #[wind_tunnel_instrument]
+    pub async fn join_space(&self) -> anyhow::Result<()> {
+        let state_lock = self.state.lock().await;
+        state_lock
+            .space
+            .local_agent_join(state_lock.agent.clone())
+            .await?;
+
+        // Wait for agent to publish their info to the bootstrap & peer store.
```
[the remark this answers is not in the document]
jost-s: No, this is the same as is done in tests in the kitsune2 repo, where the agent is joining the local space and then awaits until their info is published to the bootstrap server and peer store.

bindings/kitsune_client/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        )
+        .await
+        .unwrap();
+        let agent_1 = chatter_1.state.lock().await.agent.agent().clone();
+        let agent_2 = chatter_2.state.lock().await.agent.agent().clone();
+        chatter_1.join_space().await.unwrap();
+        chatter_2.join_space().await.unwrap();
+
+        // Bootstrapping takes about 5 seconds.
+        tokio::time::sleep(Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
jost-s: Yeup

bindings/kitsune_runner/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+bytes = { workspace = true }
+clap = { workspace = true }
+env_logger = { workspace = true }
+log = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+sha3 = { workspace = true }
+tokio = { workspace = true }
+
+kitsune2_test_utils = { git = "https://github.com/holochain/kitsune2" }
```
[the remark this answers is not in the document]
jost-s: I've used this to debug kitsune2 issues. I'll reintroduce this if needed.

scenarios/kitsune_continuous_flow/src/main.rs
```diff
@@ -0,0 +1,39 @@
+use kitsune_wind_tunnel_runner::prelude::*;
+use std::time::Duration;
+
+fn agent_setup(ctx: &mut AgentContext<KitsuneRunnerContext, KitsuneAgentContext>) -> HookResult {
+    create_chatter(ctx)?;
+    join_chatter(ctx)?;
+    ctx.runner_context().executor().execute_in_place(async {
+        tokio::time::sleep(Duration::from_secs(15)).await;
```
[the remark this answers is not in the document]
jost-s: I think it was bcause of bootstrapping. But it's not needed. Removed.

bindings/kitsune_runner/Cargo.toml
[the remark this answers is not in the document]
jost-s: https://github.com/holochain/wind-tunnel/pull/147/commits/5ba83a1935bf33f1eb566dc6d9953c598406ca4e

.github/workflows/test.yaml
jost-s: https://github.com/holochain/wind-tunnel/pull/147/commits/07b4f7afe135565ac284c1b6e4bf974add2059de

bindings/kitsune_client/src/op_store.rs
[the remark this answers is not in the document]
ThetaSinner: Yeah that never got implemented, custom metrics in the memory reporter. It's only set up to deal with timings and not other values. There's no way in general to aggregate measurements without knowing what they are and I don't really want to build a type system into the memory reporter.

I'd slightly prefer that we just run with the influx reporter if we want to see these values.

bindings/kitsune_runner/src/cli.rs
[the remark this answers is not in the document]
ThetaSinner: Nice, looks good

flake.nix
ThetaSinner: Looks like openssl went away, yay!

scenarios/kitsune_continuous_flow/Cargo.toml
ThetaSinner: Fair, in that case this seems good

ThetaSinner (approved):
Fab! One comment about reporting. I know that's messier than it should be that code! Really the only thing that works properly is the influx reporter.

ThetaSinner (approved):

PR #150, opened 2025-03-13
feat: Allow overriding `run_id` and use it as the `network_seed`

cdunster:
Being able to set the run ID allows multiple scenarios to be run with the same ID. One example is when testing a single scenario on multiple conductors with the goal of consolidating the logs into a single run.

Using the run ID as the network seed means that each run will have a unique network and will not interfere with other runs, and all conductors running a scenario as a single run will share a network seed.

### DIFF
--- a/bindings/trycp_runner/src/cli.rs
+++ b/bindings/trycp_runner/src/cli.rs
@@ -57,6 +57,12 @@ pub struct WindTunnelTryCPScenarioCli {
     /// The reporter to use.
     #[arg(long, value_enum, default_value_t = ReporterOpt::InMemory)]
     pub reporter: ReporterOpt,
+
+    /// Set the ID of this run
+    ///
+    /// If not set, a random ID is used.
+    #[arg(long, short)]
+    pub run_id: Option<String>,
 }
 
 impl TryInto<WindTunnelScenarioCli> for WindTunnelTryCPScenarioCli {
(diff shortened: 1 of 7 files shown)


bindings/trycp_runner/src/cli.rs
```diff
@@ -94,6 +94,7 @@ impl TryInto<WindTunnelScenarioCli> for WindTunnelTryCPScenarioCli {
             soak: self.soak,
             no_progress: self.no_progress,
             reporter: self.reporter,
+            run_id: None,
```
ThetaSinner, line 97: Maybe a comment that this isn't relevant for TryCP because it's centrally controlled so there's no reason to pass a run_id when starting the scenario?

ThetaSinner (approved):
Looks great! Approved with one request for a comment

bindings/trycp_runner/src/cli.rs
cdunster: Thinking about it some more, should I just add support for it for TryCP too? I guess there's no harm in adding it because people can just leave it blank to get this same functionality.

jost-s (approved):
Cool! I've done something similar for kitsune where the `run_id` is the space id. So I should make that configurable too.

jost-s (dismissed):

mattyg (approved):

bindings/trycp_runner/src/cli.rs
ThetaSinner: Yes that would make sense!

ThetaSinner (approved):

ThetaSinner (approved):

jost-s (approved):
Yeehaw

PR #151, opened 2025-03-14
chore: add missing devShell packages

cdunster:
These packages and env var are required to build.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -94,6 +94,8 @@
             pkgs.taplo
             pkgs.yamlfmt
             pkgs.perl
+            pkgs.cmake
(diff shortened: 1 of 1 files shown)


matthme (approved):

ThetaSinner (approved):

jost-s (commented):
Are those already present on the CI runners or why does it not require them?

cdunster:
> Are those already present on the CI runners or why does it not require them?

@jost-s, yeah in the CI devShell they are provided by the `pkgs.rustPlatform.bindgenHook` package which is cleaner so I updated this PR to use that package inside the default devShell too.

jost-s (approved):
Sweet

PR #152, opened 2025-03-14
Update Holochain versions and use new `AppBundleSource::Bytes` variant

cdunster:


bindings/client/src/admin_websocket.rs
```diff
@@ -25,9 +25,9 @@ pub struct AdminWebsocketInstrumented {
 impl AdminWebsocketInstrumented {
     pub async fn connect(admin_url: impl ToSocketAddr, reporter: Arc<Reporter>) -> Result<Self> {
         let addr = admin_url.to_socket_addr()?;
-        AdminWebsocket::connect(addr)
+        Ok(AdminWebsocket::connect(addr)
             .await
-            .map(|inner| Self { inner, reporter })
+            .map(|inner| Self { inner, reporter })?)
```
matthme, line 30: What's the reason for this change? Isn't it redundant?

cdunster: The error type has changed in the return type, so any errors need converting. We could use `.map_err(Into::into)` instead, but the `?` operator implicitly calls `into`.

matthme: Ah okay, nice :+1:

matthme (commented):
It looks good to me as soon as the Cargo.lock conflicts are resolved.

mattyg (approved):

jost-s (approved):

ThetaSinner (dismissed):

.github/workflows/test.yaml
```diff
@@ -75,8 +75,7 @@ jobs:
           nix flake check --all-systems
 
       - name: Run unit tests
-        run: |
-          nix develop .#ci -c bash -c "cargo test --workspace --all-targets"
+        run: cargo test --workspace --all-targets
```
cdunster, line 78: This change is required after a `nix flake update` so it seems that the latest `holonix` introduces an issue here.

jost-s: If there's nothing else needed for the tests to run, that's fine.

ThetaSinner (approved):

PR #153, opened 2025-03-14
Build out kitsune scenario

### LINKED ISSUE #146: Create a scenario that runs two instances of Kitsune2 with the above host implementation.
AC:
- Create a scenario that sets up two instances of the host implementation\
- runs a bootstrap server
- creates some ops
- asserts they have been gossiped

jost-s:
- Build out the sample scenario, add documentation and configuration options.
- Add Kitsune documentation to main README.

resolves #146

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,17 @@ The bindings contain the following crates:
 - `holochain_wind_tunnel_runner`: This is a wrapper around the `wind_tunnel_runner` crate that provides Holochain specific code to be used with the `wind_tunnel_runner`.
   The `wind_tunnel_runner` is re-exported, so you should just use this crate as your runner when creating tests for Holochain.
 
+#### Kitsune Bindings for Wind Tunnel
+
+[Kitsune](https://github.com/holochain/kitsune2) is a framework for writing distributed hash table based network applications. Bindings for Wind Tunnel enable developers
+to write scenarios to test Kitsune.
+
+The bindings contain the following crates:
+- `kitsune_client_instrumented`: A test application specifically written for Wind Tunnel. With Kitsune at its core, instances are created to publish text messages which
+are sent to all participating peers. The API is minimal, as the focus lies on observing speed and resilience of delivering messages to peers.
+- `kitsune_wind_tunnel_runner`: A wrapper around the `wind_tunnel_runner` crate that provides Kitsune specific code to be used with the `wind_tunnel_runner`. It provides
+CLI options to configure scenarios.
+
 #### Scenarios
 
 The `scenarios`, found in `./scenarios`, are what describe the performance testing scenarios to be run against Holochain. Each scenario
@@ -174,7 +185,7 @@ This is the basic structure of a Wind Tunnel scenario. The `ScenarioDefinitionBu
 a CLI which will allow you to override some of the defaults that are set in your code. Using the builder you can configure your hooks
 which are just Rust functions that take a context and return a `WindTunnelResult`. 
 
-The `run` function is then called with the builder. At that point the Wind Tunnel runner takes over and configures then runs your scenario.
+The `run` function is then called with the builder. At that point the Wind Tunnel runner takes over and configures, then runs your scenario.
 
 Before you can run this, you'll need to provide the agent behaviour hook. Add the following to your `main.rs`:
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -203,6 +214,76 @@ you are recommended to take a look at documentation for the `holochain_wind_tunn
 your scenarios and example of how to use them. This will help you get started much more quickly than starting from scratch. There is
 also a tips section below which you might find helpful as you run into questions.
 
+
+### Writing Scenarios for Kitsune
+
+> [!NOTE]
+> Writing scenarios requires some knowledge of `wind-tunnel`'s methodology as well as an overview of how Kitsune works. That is assumed knowledge for this section!
+
+Writing a Kitsune Wind Tunnel scenario is relatively straight forward. The Kitsune client defines three common functions for the developer. A chatter can be `create`d, it can `join_chatter_network` and it can `say` a list of messages. As long as a chatter has not joined the network, it won't receive messages from other peers and will also not send messages to them. Once joined, it starts receiving and sending messages it has said.
```
ThetaSinner, line 223: We are relying on gossip right? So it won't receive data from peers until it joins but it will get that data once it does join?

README.md
```diff
@@ -203,6 +214,76 @@ you are recommended to take a look at documentation for the `holochain_wind_tunn
...
+    std::thread::sleep(std::time::Duration::from_secs(1));
+    Ok(())
+}
+```
+
+This is just an example hook and you will want to replace it once you have got your scenario running. Note the `KitsuneAgentContext` that is provided
+to the hook. This is created per-agent and gives you access to the agent's ID and the runner's context as well as the chatter ID which is specific to
+Kitsune. Both the agent and the runner context are used for sharing configuration between the runner and your hooks, and state between your hooks.
+
+Your scenario should now be runnable. Try running it with
```
ThetaSinner, line 279: Will it run without a setup that joins the network?

bindings/kitsune_runner/src/common.rs
```diff
@@ -38,10 +38,13 @@ pub fn to_connection_string(bootstrap_server_url: String, signal_server_url: Str
         .expect("failed to convert bootstrap and signal server URLs to connection string")
 }
 
-/// Create a kitsune app instance.
+/// Create a Kitsune chatter instance.
 pub fn create_chatter(
     ctx: &mut AgentContext<KitsuneRunnerContext, KitsuneAgentContext>,
 ) -> HookResult {
+    if ctx.get().chatter.is_some() {
```
ThetaSinner, line 45: Nice

bindings/kitsune_runner/src/common.rs
```diff
@@ -61,26 +64,44 @@ pub fn create_chatter(
...
+        .expect("chatter_id: chatter is not created")
+        .id()
+        .to_string()
+}
+
+/// Get configured number of messages per interval.
+///
+/// This number can be configured with the env var `NUM_MESSAGES`.
+/// Defaults to 3.
+pub fn number_of_messages(ctx: &mut AgentContext<KitsuneRunnerContext, KitsuneAgentContext>) -> u8 {
```
ThetaSinner, line 81: Would slightly prefer that this was scenario-specific logic, but maybe we want to always create data at a given rate in K2 scenarios? I guess that's all we can really do anyway :D

bindings/kitsune_runner/src/definition.rs
```diff
@@ -21,6 +21,6 @@ impl<RV: UserValuesConstraint, AV: UserValuesConstraint> KitsuneScenarioDefiniti
     /// Once the Kitsune customisations have been made, use this function to switch back to
     /// configuring default properties for the scenario.
     pub fn into_std(self) -> ScenarioDefinitionBuilder<RV, AV> {
-        self.inner
+        self.inner.add_capture_env("NUM_MESSAGES")
```
ThetaSinner, line 24: Similarly this forces every scenario to use this. Maybe that's okay but I'd like to check

bindings/kitsune_runner/src/context.rs
```diff
@@ -5,15 +5,43 @@ use wind_tunnel_runner::prelude::UserValuesConstraint;
...
+            .expect(
+                "chatter is not set, did you forget to call `create_chatter` in your agent setup?",
+            )
+            .id()
+            .to_string()
+    }
+
+    /// Get configured number of messages per interval.
+    pub fn number_of_messages(&mut self) -> u8 {
+        if self.num_messages.is_none() {
```
ThetaSinner, line 38: The context is mutable, would this be clearer with a Once* type?

ThetaSinner (commented):
Really nice, I have a few comments but this is great. Especially the docs. Yay for docs :)

README.md
jost-s: Yes, I'm assuming just gossip here.

README.md
jost-s: Ah no, it won't. I was really too lazy to actually try this out!

bindings/kitsune_runner/src/common.rs
jost-s: Okay, I'll move this to just the scenario.

bindings/kitsune_runner/src/context.rs
jost-s: 'll be removed

README.md
jost-s: I've mentioned that now.

jost-s: Sorry, already squashed: https://github.com/holochain/wind-tunnel/pull/153/commits/eb28bc8b87c8d38f8e426f15c15d00a9f4336b94

bindings/kitsune_runner/src/common.rs
jost-s: https://github.com/holochain/wind-tunnel/pull/153/commits/b9e7882a5918c439791e7a9ab156f65c35b10cf0

bindings/kitsune_runner/src/definition.rs
jost-s: https://github.com/holochain/wind-tunnel/pull/153/commits/b9e7882a5918c439791e7a9ab156f65c35b10cf0

bindings/kitsune_runner/src/context.rs
jost-s: https://github.com/holochain/wind-tunnel/pull/153/commits/b9e7882a5918c439791e7a9ab156f65c35b10cf0

README.md
```diff
@@ -203,6 +214,76 @@ you are recommended to take a look at documentation for the `holochain_wind_tunn
...
+
+Then replace your `main` function with the following:
+
+```rust
+fn main() -> WindTunnelResult<()> {
+    let builder =
+        KitsuneScenarioDefinitionBuilder::<KitsuneRunnerContext, KitsuneAgentContext>::new_with_init(
+            "scenario_name",
+        )?.into_std()
+        .use_agent_behaviour(agent_behavior);
```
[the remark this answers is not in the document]
matthme: Ah, I see now that it gets defined further below :)

mattyg (approved):

README.md
ThetaSinner: Nice, yes that's good

ThetaSinner (dismissed):
Very 🧊

ThetaSinner (approved):

PR #159, opened 2025-04-03
Kitsune op store returns all processed incoming op ids

jost-s:
### Summary
- Fix WT op store to always return all processed op ids. Previously ops processed multiple times would not be removed from the request queue. Duplicate ops are still not considered for reporting.
- Add nix dev shell for kitsune scenarios.
- CI workflow uses kitsune dev shell for kitsune scenario.

### TODO:
- [x] CHANGELOGs updated with appropriate info

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -259,7 +259,7 @@ jobs:
           set -x
 
           # Start local bootstrap and signal server
-          nix develop .#ci -c bash -c "kitsune2-bootstrap-srv --listen 127.0.0.1:30000 &"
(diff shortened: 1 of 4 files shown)


ThetaSinner (dismissed):
Nice, looks good :)

We might need to add to the Kitsune shell when we integrate it with Nomad testing but still see the logic of having two shells for that

jost-s:
> Nice, looks good :)
> 
> We might need to add to the Kitsune shell when we integrate it with Nomad testing but still see the logic of having two shells for that

I wished for a separate shell when I was faced with a 4.5 GiB download just to run the kitsune scenario.

mattyg (approved):

PR #169, opened 2025-04-24
chore: Add `try_wait_for_min_peers` to Holochain runner common functions

ThetaSinner:
### TODO:

### DIFF
--- a/bindings/runner/src/common.rs
+++ b/bindings/runner/src/common.rs
@@ -11,6 +11,8 @@ use holochain_types::prelude::{
 };
 use holochain_types::websocket::AllowedOrigins;
 use std::path::PathBuf;
+use std::sync::OnceLock;
+use std::time::{Duration, Instant};
 use wind_tunnel_runner::prelude::{
     AgentContext, HookResult, RunnerContext, UserValuesConstraint, WindTunnelResult,
 };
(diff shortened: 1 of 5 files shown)


bindings/runner/src/common.rs
```diff
@@ -304,6 +306,82 @@ where
...
+/// querying peers from the conductor. You could probably not use this function for performance
+/// testing peer discovery!
+pub fn try_wait_for_min_peers<SV>(
+    ctx: &mut AgentContext<HolochainRunnerContext, HolochainAgentContext<SV>>,
+    wait_for: Duration,
+) -> HookResult
+where
+    SV: UserValuesConstraint,
+{
+    static MIN_PEERS: OnceLock<usize> = OnceLock::new();
```
cdunster, line 344: Can this not just be a normal variable?

bindings/runner/src/common.rs
```diff
@@ -304,6 +306,82 @@ where
     Ok(())
 }
 
+/// Tries to wait for a minimum number of peers to be discovered.
+///
+/// If you call this function in you agent setup then the scenario will become configurable using
+/// the `MIN_PEERS` environment variable. The default value is 2.
+///
+/// Note that the number of peers seen by each node includes itself. So having two nodes means that
+/// each node will immediately see one peer after app installation.
```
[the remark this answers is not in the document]
ThetaSinner: I'd prefer to leave it up to the user to set the right value. `+1` is only the right thing to do when you're installing one app in the scenario. It's actually `+ number of agents I know I'm creating locally`.

I have renamed this to `try_wait_for_min_agents` because that does make this clearer. I tend to use `peer` where I mean agent and they are a bit different.

jost-s (dismissed):
Wuppah! https://www.youtube.com/watch?v=ggXbzjnffAo

bindings/runner/src/common.rs
cdunster: I mean that I don't get why this is a `OnceLock`. Maybe I'm missing something but AFAIK it would work without the static and changing the `let min_agents = ...` line a couple of lines below to be:

```rust
let min_agents = std::env::var("MIN_AGENTS")
    .ok()
    .map(|s| s.parse().expect("MIN_AGENTS must be a number"))
    .unwrap_or(2);
```

I don't think the `OnceLock` adds anything?

ThetaSinner: I get you. Because this is only meant to get called once I agree. The intention was to cache the calculation so that it can be called more than once and get the same result but that's not what this is for.

https://github.com/holochain/wind-tunnel/pull/169/commits/be5154f8a74db6adf5509b9519efc0a8b585f087

cdunster (approved):

PR #185, opened 2025-05-20
ci: Use less disk space

ThetaSinner:
### TODO:


.github/workflows/test.yaml
```diff
@@ -40,11 +40,20 @@ jobs:
 
       - uses: cachix/cachix-action@v15
         with:
-          name: holochain-wind-tunnel
-          authToken: ${{ secrets.CACHIX_HOLOCHAIN_WIND_TUNNEL }}
+          name: holochain-ci
```
ThetaSinner, line 43: Switched to using our bigger cache. The `holochain-wind-tunnel` one was a free account and only had 5GB. That's part of the problem here.

.github/workflows/test.yaml
```diff
@@ -71,8 +80,7 @@ jobs:
 
       - name: Lint Rust
         run: |
-          # Currently the only check is clippy, could bundle the other checks into the flake and remove the steps above?
-          nix flake check --all-systems
```
ThetaSinner, line 75: This all system check was downloading things for other systems and I don't think it's ever caught an actual problem. We test on linux and macos, that's covering everything we're actually using.

.github/workflows/test.yaml
```diff
@@ -253,17 +261,45 @@ jobs:
           # Stop servers
           pkill kitsune2-bootst
 
+  archive_bundles:
```
ThetaSinner, line 264: Split into a separate workflow. You can think of this as:

- Everything useful from the previous step was cached on Cachix.
- All the Rust caching and Nix store that was build in the last job is dropped by switching machine.
- We can download anything that's useful for this job from Cachix.

This actually has to download a bunch of software that the previous steps haven't needed too, so it seemed best to have it be separate.

I have kept a `needs:` here because we don't want to bundle the scenarios and archive them if they're not working!

flake.nix
```diff
@@ -149,7 +149,7 @@
   });
 
   nixConfig = {
-    substituters = [ "https://cache.nixos.org" "https://holochain-ci.cachix.org" "https://holochain-wind-tunnel.cachix.org" ];
```
ThetaSinner, line 152: No longer need the wind tunnel cache

scripts/exec_on_targets.ab
ThetaSinner, line 1: We don't use this any more, it was for running things on the list of TryCP targets. It also looks like Amber is really slow to build in CI

flake.nix
```diff
@@ -7,20 +7,10 @@
 
     holonix = {
       url = "github:holochain/holonix?ref=main-0.4";
-      inputs = {
```
ThetaSinner, line 10: This is something that has been presented to me by the people who knew more about Nix than me as a good thing to do. So that you only build with one copy of nixpkgs and other inputs. Sounds great, in reality, it breaks caching all the time.

I noticed that some things that are definitely cached on Cachix by the Holonix CI are being rebuilt here.

.github/workflows/test.yaml
```diff
@@ -89,7 +93,7 @@ jobs:
           nix develop .#ci --command bash ./scripts/wait-for-hc.sh
 
           # Run the scenario
-          RUST_LOG=info nix run .#zome_call_single_value -- --connection-string ws://localhost:8888 --duration 5 --no-progress
+          RUST_LOG=info cargo run --package zome_call_single_value -- --connection-string ws://localhost:8888 --duration 5 --no-progress
```
ThetaSinner, line 96: As people had added to this workflow, a mixture of `cargo` and `nix` was being used. That meant we were building all the project dependencies twice and some of the test zomes twice. Just switched to cargo since we're running the tests with cargo anyway

jost-s (approved):
Awesome! I've found the workflow endless and especially the Lint Rust step takes a long time and nothing was cached when it failed.

PR #198, opened 2025-06-18
chore: Use workspace package properties

ThetaSinner:
### TODO:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,9 +147,7 @@ dependencies = [
 name = "app_install"
 version = "0.1.0"
 dependencies = [
- "anyhow",
  "happ_builder",
- "holochain_types",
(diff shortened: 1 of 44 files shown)


ThetaSinner:
> I do wonder if the happ_builder and sumeriser crates should also use the workspace version?

That could be done, but they're both so opinionated that I don't see a good reason to publish them to crates.io. If somebody used the WT libraries for something, they wouldn't necessarily find those tools useful. Even the hApp builder, I wouldn't expect anyone else to be managing a library of test hApps, but instead be testing against an existing hApp that they're trying to validate performance for. I think if we did publish them, then it should be with the workspace version - but while they're `publish = false`, I would slightly prefer not to have them versioned.

PR #215, opened 2025-07-03
feat: Add scenario showing how to use paths

cdunster:
### Summary

As part of [the path validation spike](https://github.com/holochain/holochain/issues/3178) I wanted to explore the ways that paths could be used in their current form. I decided to do it as a Wind Tunnel scenario so that it is minimal and self-standing. I hope that this can be used as an example of using paths and validating them.

### TODO:
- [ ] Remove old commits with code that went nowhere
- [ ] Add support for running with Nomad
- [ ] Separate the agent behaviour to have one (or more?) agents that create different entries and check that they can see each other's.

### DIFF
--- a/zomes/path_validated/integrity/src/lib.rs
+++ b/zomes/path_validated/integrity/src/lib.rs
@@ -30,18 +30,9 @@ fn validate(op: Op) -> ExternResult<ValidateCallbackResult> {
             tag,
             ..
         }) => {
-            let tag_bytes = SerializedBytes::from(UnsafeBytes::from(tag.into_inner()));
-            let tag_component = Component::try_from(tag_bytes).map_err(|e| wasm_error!(e))?;
-            let tag_string = String::try_from(&tag_component).map_err(|e| wasm_error!(e))?;
+            let tag_bytes = SerializedBytes::from(UnsafeBytes::from(tag.clone().into_inner()));
 
-            if tag_string
-                .chars()
-                .any(|c| c != '-' && !c.is_ascii_lowercase())
-            {
-                Ok(ValidateCallbackResult::Invalid(format!(
-                    "Link's tag of '{tag_string:?}' contained more than lower-case ASCII letters and dashes",
-                )))
-            } else if base_address.clone().into_entry_hash().is_none() {
+            if base_address.clone().into_entry_hash().is_none() {
                 Ok(ValidateCallbackResult::Invalid(format!(
                     "Link's base_address '{base_address}' was not a valid entry hash",
                 )))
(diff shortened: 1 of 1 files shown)


zomes/path_validated/coordinator/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+fn recursively_create_links_from_root(
+    path: TypedPath,
+    book_entry_hash: EntryHash,
+) -> ExternResult<()> {
+    let parent_hash = if let Some(parent) = path.parent() {
+        recursively_create_links_from_root(parent.clone(), book_entry_hash.clone())?;
+        parent.path_entry_hash()?.into()
+    } else {
+        root_hash()?
```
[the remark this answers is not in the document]
matthme: As I understand it, one of the reasons for using paths is to distribute the load across nodes since each path has a different hash and consequently a different authority (modulo randomness) is responsible for storing links pointing away from it . So I think we actually don't want to ever create a link from the root hash or else it defeats the purpose of reducing DHT hot-spots in that way, if all entries ultimately always have a link to the root hash. In fact, it makes me wonder whether entries should in most cases only be linked to a leaf of the path structure.

zomes/path_validated/coordinator/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    .is_empty()
+    {
+        // Link to book with name as tag exists so the book should exist.
+        return Ok(());
+    }
+
+    let book_action_hash = create_entry(EntryTypes::BookEntry(BookEntry {
+        name: author_and_name.1,
+    }))?;
+    let book_action = must_get_action(book_action_hash)?;
```
ThetaSinner, line 36: `must_get_*` in a coordinator? I'd expect this to just be a `get`?

zomes/path_validated/integrity/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+            if TryInto::<String>::try_into(tag.clone()).is_err() {
+                Ok(ValidateCallbackResult::Invalid(format!(
+                    "Link's tag of '{tag:?}' was not a valid string",
+                )))
+            } else if base_address.clone().into_entry_hash().is_none() {
+                Ok(ValidateCallbackResult::Invalid(format!(
+                    "Link's base_address '{base_address}' was not a valid entry hash",
+                )))
+            } else if let Some(book_entry_hash) = target_address.clone().into_entry_hash() {
+                if must_get_entry(book_entry_hash).is_ok() {
```
ThetaSinner, line 82: Possibly could be a little stricter here? Rather than it just existing:
- Retrieve it and check the entry type
- Based on the content, you could construct the author path and verify that the base address of this link matches the entry hash of the path

ThetaSinner (commented):
Test logic and zomes look good, just minor comments there.

For the scenario though, it really reads to me like a functional test and not a performance test. I think we could have both but:
- I'd like to transplant most of this code to `holochain/holochain` and have a functional test for path validation there.
- As a scenario that does interesting validation and exercises a real use-case, I'm happy with this being here. But to make this more of a scale/performance scenario it would have to keep creating new data to be validated and not used fixed test data that is testing the correctness of the behavior. We'd also need to figure out what metrics to record.

PR #219, opened 2025-07-14
ci: add missing CACHIX_AUTH_TOKEN env to cachix push step

cdunster:
### Summary

The cachix token is not present in the push job so it fails: https://github.com/holochain/wind-tunnel/actions/runs/16260690774/job/45905106311

I think that it's previously passed because there has been nothing to push so it doesn't try.

### TODO:
- [x] CHANGELOGs updated with appropriate info


ThetaSinner (approved):
Well yeah that wouldn't work 😕

PR #220, opened 2025-07-15
chore: add cargo clean step to Rust static check

cdunster:
### TODO:

### DIFF
--- a/scripts/checks.sh
+++ b/scripts/checks.sh
@@ -17,6 +17,7 @@ check_rust_fmt() {
 }
 
 check_rust_static() {
+  cargo clean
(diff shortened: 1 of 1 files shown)


jost-s (approved):
So this is the solution after all?!

cdunster:
> So this is the solution after all?!

@jost-s, I'm not sure because [my other PR](https://github.com/holochain/wind-tunnel/pull/221) just passed without these changes, so maybe it was random :confounded:


> Is this needed because we're mixing builds between Nix and `cargo` still?

@ThetaSinner, maybe. Maybe because we enter the `devShell` and run the checks, then the results are cached in the `devShell` derivation, but this doesn't sound right to me and as mentioned above, the other PR passed without this...

PR #233, opened 2025-07-29
Fix generated nomad jobs and CI that generates them

cdunster:
### TODO:

## Summary by CodeRabbit

* **Chores**
  * Improved workflow to dynamically select job parameters during Nomad job builds.
  * Updated Nomad job configuration to change the source location for the `INFLUX_TOKEN` environment variable retrieval.

### DIFF
--- a/nomad/run_scenario.tpl.hcl
+++ b/nomad/run_scenario.tpl.hcl
@@ -168,7 +168,7 @@ job "{{ (ds "vars").scenario_name }}" {
             destination = "${NOMAD_SECRETS_DIR}/secrets.env"
             env         = true
             data        = <<EOT
-            INFLUX_TOKEN={{ "{{" }} with nomadVar "nomad/jobs/{{ (ds "vars").scenario_name }}" {{ "}}" }}{{ "{{" }} .INFLUX_TOKEN {{ "}}" }}{{ "{{" }} end {{ "}}" }}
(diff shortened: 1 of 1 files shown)


nomad/run_scenario.tpl.hcl
```diff
@@ -168,7 +168,7 @@ job "{{ (ds "vars").scenario_name }}" {
             destination = "${NOMAD_SECRETS_DIR}/secrets.env"
             env         = true
             data        = <<EOT
-            INFLUX_TOKEN={{ "{{" }} with nomadVar "nomad/jobs/{{ (ds "vars").scenario_name }}" {{ "}}" }}{{ "{{" }} .INFLUX_TOKEN {{ "}}" }}{{ "{{" }} end {{ "}}" }}
+            INFLUX_TOKEN={{ with nomadVar "nomad/jobs" }}{{ .INFLUX_TOKEN }}{{ end }}
```
cdunster, line 171: This allows us to have a single token that is accessable to all Nomad jobs instead of needed to generate a new token for every job and adding it to the Nomad cluster.

See: https://nomad-server-01.holochain.org:4646/ui/variables/var/nomad/jobs/run_scenario@default

ThetaSinner (dismissed):

nomad/run_scenario.tpl.hcl
cdunster: @ThetaSinner, I forgot to escape this so gomplate thought it was a gomplate thing instead of a Nomad thing. I've amended it now.

ThetaSinner (approved):

PR #235, opened 2025-07-30
feat: Integrate Host Metrics collection with reporting

veeso:
### Summary

- Added a PostRun worker to collect metrics after scenario run
- Implemented a HostMetrics colletor and reporter to collect HostMetrics from Telegraf output file

### TODO:


framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,99 @@
...
+                    log::error!("Failed to read line from metrics file: {e}");
+                    continue;
+                }
+            };
+            // parse line into metrics
+            let metrics = match HostMetrics::from_str(&line) {
+                Ok(metrics) => metrics,
+                Err(e) => {
+                    log::error!("Failed to parse metrics line '{line}': {e}");
+                    continue;
```
veeso, line 60: I'm not sure here what to do. Should we fail? Or ignore and report the error?

framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,152 @@
...
+                }
+            };
+            // parse line into metrics
+            let metrics = match HostMetrics::from_str(&line) {
+                Ok(metrics) => metrics,
+                Err(e) => {
+                    log::error!("Failed to parse metrics line '{line}': {e}");
+                    errors += 1;
+                    continue;
+                }
```
veeso, line 66: Do we have to filter by run_id here?

framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,152 @@
...
+            .map_err(|e| anyhow::anyhow!("Failed to open metrics file: {e}"))?;
+
+        let mut errors = 0;
+        let mut reported = 0;
+
+        // iterate over lines
+        let reader = BufReader::new(file);
+        for line in reader.lines() {
+            // handle each line, expecting it to be a valid HostMetrics entry
+            // we don't return errors for each line, but log them instead
```
matthme, line 49:
```suggestion
            // we don't return errors for each line, but log them instead.
```

framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,152 @@
...
+            };
+
+            self.reporter.add_custom(metrics.report_metric());
+            reported += 1;
+        }
+
+        log::info!("Reported {reported} host metrics with {errors} errors.");
+
+        // if we have no valid metrics reported but encountered errors,
+        // we return an error to indicate that something went wrong
```
matthme, line 76:
```suggestion
        // we return an error to indicate that something went wrong.
```

framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[cfg(test)]
+mod tests {
+
+    use crate::test_utils::test_reporter;
+
+    use super::*;
+
+    const JSON_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/tests/host_metrics.json");
+    const MALFORMED_JSON_PATH: &str = concat!(
```
matthme, line 97:
```suggestion
    const MALFORMED_JSON_PATH_ZERO_VALID_METRICS: &str = concat!(
```

framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,152 @@
...
+        env!("CARGO_MANIFEST_DIR"),
+        "/tests/malformed_host_metrics.json"
+    );
+    const MALFORMED_JSON_WITH_ONE_VALID_PATH: &str = concat!(
```
matthme, line 101:
```suggestion
    const MALFORMED_JSON_WITH_ONE_VALID_METRIC: &str = concat!(
```

framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,152 @@
...
+    async fn test_should_report_host_metrics() {
+        let reporter = test_reporter();
+        let metrics_file = PathBuf::from(JSON_PATH);
+        let host_metrics_reporter = HostMetricsReporter::new(metrics_file, reporter.clone());
+
+        assert!(host_metrics_reporter.report().is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_should_report_error_if_malformed_file() {
```
matthme, line 118:
```suggestion
    async fn test_should_report_error_if_zero_valid_metrics() {
```

framework/runner/src/postrun/host.rs
```diff
@@ -0,0 +1,152 @@
...
+        let reporter = test_reporter();
+        let metrics_file = PathBuf::from(MALFORMED_JSON_PATH);
```
matthme, line 120:
```suggestion
        let metrics_file = PathBuf::from(MALFORMED_JSON_PATH_ZERO_VALID_METRICS);
```

ddd-mtl:
It seems we expect the host metric file to be json only. Maybe this expectation should be make explicit in the CLI doc. Would we also want to handle line protocol files?
I'd suggest checking the host metric file validity and report any error before running the tests. In the current PR it seems the error will be reported after the run.

metrics/import_host_metrics/src/aggregator.rs
```diff
@@ -0,0 +1,182 @@
...
+            .filter(|m| m.timestamp >= start_time && m.timestamp <= end_time)
+            .collect()
+    }
+
+    /// Create a [`WriteQuery`] from its relative [`RunScenario`], `host_id` and the [`HostMetrics`] entry.
+    fn create_report(run_scenario: &RunScenario, host_metrics: &HostMetrics) -> WriteQuery {
+        // NOTE: keep nanoseconds, otherwise it won't work with InfluxDB
+        let metric = WriteQuery::new(
+            influxdb::Timestamp::Nanoseconds(host_metrics.timestamp as u128 * 1_000_000_000),
+            format!("wt.host.{}", host_metrics.name),
```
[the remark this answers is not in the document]
veeso: We don't assume, it's in the docs.

metrics/import_host_metrics/src/metrics/fields.rs
```diff
@@ -0,0 +1,590 @@
+use influxdb::WriteQuery;
```
veeso, line 1: I've restored the census of all the fields. Unfortunately it is necessary, because we need always to know whether it's an integer or a float for influx.

I've tried to do a as_u64 and as_f64 on json values, but it doesn't work well, because if a float has not decimal when parsed, it is treated as an integer and influx fails when you import different types for the same values.

So probably the best way is to do like this.

ThetaSinner (commented):
As meta comments:
- I am disappointed that my companion PR wasn't given a more detailed consideration. For example, the "tags" default plugin won't be working on this PR and I had pointed that out on my PR. That the problem isn't resolved here makes me think that either some configuration just got left behind in the code or my PR wasn't read carefully. Even if it's just left behind, because the tool deals with setting default values, I would hope that my comments would have sparked checking that on this PR.
- Any changes that are being made here need to be updated in the documentation. Although you have raised that the documentation could use work, it does contain all the things that can be done, it's really just missing a coherent description of how to combine the documented steps. This PR only comes with documentation for the tool and not documentation changes for how we're supposed to run scenarios. Even if that is just additions to suggest that the tool needs to be used for getting host metrics and pointing to the tool's own README.
- I have done a detailed review to see if I could adjust to your suggestion that the tool was a better option than my suggestion. I have not been convinced and I think the tool is well-coded but ultimately re-implementing what Telegraf does. The unique piece is the timestamp logic but see my comments below about why that is less important - I think it's unnecessary on Nomad and locally it's acceptable to need a couple of extra steps to get host metrics for your scenario. We need that filtering on timestamps later in the pipeline, and the summarizer tool is best equipped to handle that. I am still taking the position that tooling added for this issue should be glue code that automates other tools and doesn't otherwise do work itself.

> The tool now stands at 700 CLOCs and is well-tested. 

I agree that the tool is well-tested. Having done a thorough review though, the tool is mostly re-implementing what Telegraf can already do. 

> I would say this tool is easier to maintain (and probably won't require any maintenance ...)

I disagree with this, because I already found something that this tool doesn't do that Telegraf does which is batching to limit memory requirements. I would also recommend that we don't use the library that has been picked for working with Influx line data. The least maintenance we can do is not interact with the line protocol at all and let existing tools process it. 

> Additionally, with the tool, we're confident that it handles edge cases more effectively. I also believe it should retrieve not just the latest scenario, but all scenarios in the summary.

I assert that it does not need to handle edge cases. Edge cases that Telegraf can't handle should generally be acceptable failures. The filter I wrote for picking the most recent run_id out of the run_summary.jsonl was an affordance for running locally and within a Nomad job there would just be one run in existence because each job should get a clean working directory.

Locally, host metrics are almost always going to be meaningless. There are a couple of scenarios where we might be interested, such as the install-app scenario, but in general we care about multi-user networks and running locally won't produce interesting host metrics. I think the only time where we'd really be collecting some host metrics is when creating test data for the summarizer tool. Otherwise, we really only need them when scenarios are being run on Nomad. I wouldn't use them locally myself if I didn't need them for something specific.

metrics/import_host_metrics/src/jsonl.rs
```diff
@@ -0,0 +1,88 @@
+use std::io::BufRead as _;
+
+/// A module for reading JSON Lines (jsonl) files.
+#[derive(Debug, Default)]
+pub struct JsonlReader {
```
ThetaSinner, line 5: There is a `load_run_summary` function in the summary_model that does something similar to this

Cargo.toml
```diff
@@ -87,6 +89,7 @@ parking_lot = "0.12.1"
 log = "0.4.20"
 env_logger = "0.11.2"
 derive_more = { version = "2", features = ["error", "display", "debug"] }
+influxlp-tools = "0.2.3"
```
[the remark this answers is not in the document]
veeso: Libs wno updates are not dead projects, nor unreliable; it's just a completed library, which does what it must do. 
We are using a tool that has no official support for Rust, so we have to rely on something else.
There used to be a popular library for influx. Still, it was developed by a very famous individual who has a terrible reputation for deleting projects from the internet when he's tired of supporting them.

PR #237, opened 2025-08-05
fix: Fixed host dashboards to skip custom buckets

veeso:
### Summary

There was an issue which was not seen when importing the template from influx, but only when setting up with nix command, that the variables were not created.

We actually don't need those variables on the host metrics, since we always use the same bucket and we don't collect for more hosts.

### TODO:


telegraf/telegraf.conf
```diff
@@ -133,6 +133,33 @@
   ## By default, telegraf will gather stats for all devices including
   ## disk partitions.
 
+[[inputs.file]]
```
veeso, line 136: Also restored this to import metrics from influx files.

veeso:
I've already updated our server

veeso:
@ThetaSinner 

All the queries in the template had `from(bucket: v.bucket)` and this didn't work because it says we don't have this variable, instead we should just have `from(bucket "windtunnel")`.

Also, later it applied a filter to the `linux_host`, which is an unexisting variable and we don't need it actually, so I removed it as well.

matthme:
Is there a reason for these json files not to be formatted in a readable manner in which diffs can also be spotted clearly on PRs? Are they auto-generated or something like that?

veeso:
@matthme I don't know, none of them is formatted, so I kept as-is, but yeah it could make sense to format. They are not auto generated, they are downloaded from influx UI, but not automatically

matthme:
Okay. I think it would be worth to auto-format them then, maybe as a separate PR.

veeso:
@matthme I'll format them in a separate PR, merge it, rebase this on that and so the change here is more clear

veeso:
https://github.com/holochain/wind-tunnel/pull/239

ThetaSinner (approved):

ThetaSinner (approved):

PR #241, opened 2025-08-07
feat: Collect host metrics

ThetaSinner:
### TODO:

cpu-with-run-id
dht-sync-lag-run-id

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,7 +2,7 @@
   description = "Flake for Holochain testing";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-25.05";
+    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
     flake-parts.url = "github:hercules-ci/flake-parts";
 
     holonix = {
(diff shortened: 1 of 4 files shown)


telegraf/host.conf
ThetaSinner, line 1: `host.conf` runs an agent, uses system metrics as inputs and writes content to the same place that scenario metrics are written

flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "Flake for Holochain testing";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-25.05";
+    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
```
ThetaSinner, line 5: The version of Telegraf available in 25.05 is 1.34 and doesn't support adding default tags, only fields. Don't know what else I've imported at newer versions here, just wanted a newer Telegraf binary at 1.35 or later.

flake.nix
```diff
@@ -111,10 +111,35 @@
         packages = {
           default = config.workspace.workspace;
           inherit (config.workspace) workspace;
+          local-telegraf = pkgs.writeShellApplication {
+            name = "local-telegraf";
+            runtimeInputs = [ pkgs.telegraf ];
+            text = ''
+              RUN_ID=$(jq --slurp --raw-output 'sort_by(.started_at|tonumber) | last | .run_id' < run_summary.jsonl)
+
+              sed --in-place "s/run_id = \"\"/run_id = \"$RUN_ID\"/" ./telegraf/local-telegraf.conf
```
ThetaSinner, line 120: Put the latest run_id in the telegraf config

flake.nix
```diff
@@ -111,10 +111,35 @@
...
+
+              echo "Running telegraf for run ID: $RUN_ID"
+              telegraf --config telegraf/local-telegraf.conf --once > >(tee logs/telegraf-stdout.log) 2> >(tee logs/telegraf-stderr.log >&2)
+
+              rm ./telegraf/metrics/*.influx
```
ThetaSinner, line 125: Remove any metrics that have now been uploaded. Don't want to re-use/relabel any old host metrics.

flake.nix
```diff
@@ -111,10 +111,35 @@
...
+              git checkout -- telegraf/local-telegraf.conf
+            '';
+          };
           ci-telegraf = pkgs.writeShellApplication {
             name = "ci-telegraf";
             runtimeInputs = [ pkgs.telegraf ];
-            text = "telegraf --config telegraf/runner-telegraf.conf --once > >(tee logs/telegraf-stdout.log) 2> >(tee logs/telegraf-stderr.log >&2)";
+            text = ''
```
ThetaSinner, line 132: Identical steps, just working against the runner telegraf config with a different influx instance configured

PR #245, opened 2025-08-11
ci: Wait for Nomad CI jobs to run

### LINKED ISSUE #164: Create a CI job that reports the result of the scenarios run in Nomad
Currently, the Nomad CI job will succeed if it successfully places the Nomad jobs. This means that the CI does not report the actual results of running the scenarios. It would be nice to wait on the results of the scenarios running in the Nomad cluster and report that in the CI. Nomad does not seem to natively support waiting for the job to complete, however, we can poll the status of the running allocations and see the result of them. Therefore, we can collect the allocation ID for each of the jobs that are running and have a final CI step or job that waits for the Nomad job to complete and report the result.

See the Nomad docs for the [`alloc status` command](https://developer.hashicorp.com/nomad/docs/commands/alloc/status)

veeso:
closes #164

### Summary

- Added a step to Nomad ci jobs to wait for nomad jobs to complete

### TODO:

## Summary by CodeRabbit

* **New Features**
  * CI now captures Nomad job output and allocation IDs, then waits for allocations to finish, making submit-then-wait the default flow and surfacing allocation status in step summaries.

* **Chores**
  * Fails early and reports when no allocation is created, adds a sensible timeout, validates required environment and tools, supports sequential handling of multiple allocations, and emits concise run summaries for clearer outcomes.

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -89,6 +89,7 @@ jobs:
           echo "download-url=$DOWNLOAD_URL" >> "$GITHUB_OUTPUT"
 
       - name: Run Nomad Job
+        id: run-nomad-job
         env:
           NIXPKGS_ALLOW_UNFREE: 1
           NOMAD_ADDR: https://nomad-server-01.holochain.org:4646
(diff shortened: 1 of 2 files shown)


.github/workflows/nomad.yaml
```diff
@@ -97,6 +98,22 @@ jobs:
           NOMAD_VAR_scenario-url: ${{ steps.get-download-url.outputs.download-url }}
           NOMAD_VAR_run-id: ${{ github.run_id }}
         run: |-
-          nix run --impure --inputs-from . nixpkgs#nomad -- job run nomad/jobs/${{ matrix.job-name || matrix.scenario-name }}.nomad.hcl
+          set -Eeuo pipefail
```
ThetaSinner, line 101: Is `-E` needed? That's not one we normally set

nomad/wait_for_jobs.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+
+# Timeout in seconds (30 minutes). Can be overridden via environment.
+TIMEOUT="${TIMEOUT:-1800}"
+
+set -Eeuo pipefail
```
ThetaSinner, line 6:
```suggestion
set -Eeuo pipefail

# Timeout in seconds (30 minutes). Can be overridden via environment.
TIMEOUT="${TIMEOUT:-1800}"
```

ThetaSinner (approved):
Nice!

.github/workflows/nomad.yaml
veeso: Removed

ThetaSinner (approved):

PR #250, opened 2025-08-14
feat(metrics): Integrated Host metrics into Nomad runs

### LINKED ISSUE #210: Integrate host metrics with Nomad
As part of the parent feature, host metrics will be possible to push to a development instance of InfluxDB with a `run_id` associated with the scenario that ran. Now that needs to happen when we run scenarios using the Nomad environment and the metrics need to end up in the live InfluxDB instance.

AC:
- When a scenario is run with Nomad, then host metrics are produced to InfluxDB along with the other metrics we're already capturing.

veeso:
### Summary

Also removed sed and temp telegraf config; read RUN_ID from env (or default to "").

closes #210

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Host-level metrics collector added as a prestart sidecar and a post-run metrics upload step.

* **Documentation**
  * Fixed README example URL (http::// → http://).

* **Refactor**
  * Run ID and run-summary path are now environment-driven; metric configs and output locations read from environment rather than per-run file mutation.

* **Chores**
  * Improved job-wait diagnostics and consolidated task log extraction; CI run-id wiring updated.


flake.nix
```diff
@@ -138,15 +138,10 @@
               else
                 RUN_ID=""
               fi
-
-              # get tempfile for telegraf config
-              TELEGRAF_CONFIG=$(mktemp --suffix ".conf")
-              cp ./telegraf/telegraf.local.conf "$TELEGRAF_CONFIG"
-
-              sed --in-place "s/run_id = \"\"/run_id = \"$RUN_ID\"/" "$TELEGRAF_CONFIG"
+              export RUN_ID
```
veeso, line 141: We can set RUN_ID in the telegraf config via env instead of creating a temp config

telegraf/telegraf.local.conf
```diff
@@ -79,4 +79,4 @@
   ## more spaces.
   ## <target-tag> = <value>
   [processors.defaults.tags]
-    run_id = ""
+    run_id = "${RUN_ID-}"
```
veeso, line 82: run from env and default to empty

nomad/run_scenario.tpl.hcl
```diff
@@ -98,6 +98,28 @@ job "{{ (ds "vars").scenario_name }}" {
         }
       }
 
+      task "report_host_metrics" {
+        lifecycle {
+          hook = "prestart"
+          sidecar = true
+        }
+
```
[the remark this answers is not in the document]
cdunster: Note that the `upload_metrics` task will only run if `reporter` is set to `"influx-file"` so if it's not then I guess we'll gather the metrics and not upload them. Maybe it's good to change the `upload_metrics` task to always run so we always have the host metrics at least?

nomad/run_scenario.tpl.hcl
```diff
@@ -162,6 +184,7 @@ job "{{ (ds "vars").scenario_name }}" {
           env {
             TELEGRAF_CONFIG_PATH = "${NOMAD_TASK_DIR}/telegraf.runner.conf"
             WT_METRICS_DIR       = "${NOMAD_ALLOC_DIR}/data/telegraf/metrics"
+            RUN_ID               = "${var.run-id != null ? var.run-id : ""}"
```
cdunster, line 187: So if we don't set `run-id` in Nomad then the `RUN_ID` env is not set, meaning that we collect the host metrics but don't associate them with a particular run? And then the scenario will still run with a random ID, right?

nomad/upload_metrics.sh
```diff
@@ -0,0 +1,37 @@
...
+  local var_name="$1"
+  # temporarily allow fail to bind
+  set +u
+  if [[ -z "${!var_name}" ]]; then
+    echo "Environment variable $var_name is not set." >&2
+    set -u
+    exit 1
+  fi
+  set -u
+}
```
cdunster, line 18: Couldn't we just first check if it's set so that we don't have to use `set +u/-u`?

```suggestion
function check_envset() {
  local var_name="$1"
  if [[ ! -v "$var_name" ]] || [[ -z "${!var_name}" ]]; then
    echo "Environment variable $var_name is not set or is an empty string." >&2
    exit 1
  fi
}
```

nomad/upload_metrics.sh
```diff
@@ -0,0 +1,37 @@
...
+
+check_envset "WT_METRICS_DIR"
+check_envset "INFLUX_TOKEN"
+check_envset "TELEGRAF_CONFIG_PATH"
+
+# read RUN_ID if not set in the environment
+if [ -z "${RUN_ID-}" ]; then
```
cdunster, line 25: Couldn't this also use `-v`?
```suggestion
if [ -v "$RUN_ID" ]; then
```

nomad/upload_metrics.sh
```diff
@@ -0,0 +1,37 @@
...
+    summary_file="${WT_METRICS_DIR}/run_summary.jsonl"
+    if [ -f "$summary_file" ]; then
+        RUN_ID=$(jq --slurp --raw-output 'sort_by(.started_at|tonumber) | last | .run_id' < "$summary_file")
```
cdunster, line 28: Will this work? I'm don't know if `jq` is installed on the Nomad clients?

nomad/wait_for_jobs.sh
```diff
@@ -7,10 +7,14 @@ TIMEOUT="${TIMEOUT:-1800}"
...
 function check_envset() {
   local var_name="$1"
+  # temporarily allow fail to bind
+  set +u
   if [[ -z "${!var_name}" ]]; then
     echo "Environment variable $var_name is not set." >&2
+    set -u
     exit 1
   fi
+  set -u
```
cdunster, line 17:
```suggestion
  if [[ ! -v "$var_name" ]] || [[ -z "${!var_name}" ]]; then
    echo "Environment variable $var_name is not set or is an empty string." >&2
    exit 1
  fi
```

nomad/upload_metrics.sh
veeso: @cdunster I've reverted. It would be good, but I've just found out, just because I was testing another issue on macos, that on MacOS bash 3.2 is installed, which doesn't support -v.

nomad/run_scenario.tpl.hcl
```diff
@@ -123,10 +152,11 @@ job "{{ (ds "vars").scenario_name }}" {
...
         env {
-          RUST_LOG       = "info"
-          HOME           = "${NOMAD_TASK_DIR}"
-          WT_METRICS_DIR = "${NOMAD_ALLOC_DIR}/data/telegraf/metrics"
-          MIN_AGENTS     = "${var.min-agents}"
+          RUST_LOG          = "info"
+          HOME              = "${NOMAD_TASK_DIR}"
+          WT_METRICS_DIR    = "${NOMAD_ALLOC_DIR}/data/telegraf/metrics"
+          MIN_AGENTS        = "${var.min-agents}"
+          RUN_SUMMARY_PATH  = "${NOMAD_ALLOC_DIR}/run_summary.jsonl"
```
veeso, line 159: path for reporting run summary

nomad/upload_metrics.sh
```diff
@@ -0,0 +1,45 @@
...
+if [ -z "$RUN_ID" ]; then
+    # if isset RUN_SUMMARY_PATH
+    summary_path="run_summary.jsonl"
+    if [ ! -z "$RUN_SUMMARY_PATH" ]; then
+        summary_path="$RUN_SUMMARY_PATH"
+    fi
+
+    if [ -f "$summary_path" ]; then
+        # NOTE: nomad workspace doesn't have jq installed, so we can't use it
+        RUN_ID=$(grep -o '"run_id":"[^"]*"' "$summary_path" | sed 's/.*:"\([^"]*\)".*/\1/' | tail -n 1)
```
veeso, line 35: jq indeed doesn't work on nomad. This is the only command combination I could find.

framework/runner/src/run.rs
```diff
@@ -244,7 +250,13 @@ pub fn run<RV: UserValuesConstraint, V: UserValuesConstraint>(
     runner_context_for_teardown.reporter().finalize();
 
     summary.set_peer_end_count(agents_run_to_completion.load(std::sync::atomic::Ordering::Acquire));
-    if let Err(e) = append_run_summary(summary, PathBuf::from("run_summary.jsonl")) {
+
+    // append run summary
+    let summary_path = std::env::var(RUN_SUMMARY_PATH_ENV)
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| PathBuf::from(DEFAULT_RUN_SUMMARY_PATH));
```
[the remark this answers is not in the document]
veeso: But you can't map an unwrapped value.

nomad/upload_metrics.sh
```diff
@@ -0,0 +1,44 @@
...
+  local var_name="$1"
+  # temporarily allow fail to bind
+  # Do not use `[[ -v $var_name ]]` as it is not compatible with MacOS bash (still using 3.2)
+  set +u
+  if [[ -z "${!var_name}" ]]; then
+    set -u
+    echo "Environment variable $var_name is not set or is an empty string." >&2
+    exit 1
+  fi
+  set -u
```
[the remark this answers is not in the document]
veeso: this works actually:

```bash
function check_envset() {
  local var_name="$1"
  if [[ -z "${!var_name:-}" ]]; then
    echo "Environment variable $var_name is not set or is an empty string." >&2
    exit 1
  fi
}
```

veeso:
I've tested jq on the runner and it works now.

framework/runner/src/run.rs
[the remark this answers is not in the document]
cdunster: Yeah, it's fine. I meant to use `into`, but then I think you'd have to do `to_string` on the default value anyway, and supply the type, so what you have is good :slightly_smiling_face:

cdunster (approved):

Issue #204, opened 2025-06-30
Document collecting Holochain metrics in a dev environment


ThetaSinner:
In a dev environment (Nix shell), we should be able to configure Holochain to report metrics to a development instances of InfluxDB.

AC:
- README updated with instructions for configuring Holochain to report metrics to InfluxDB.
- Metrics show up in InfluxDB with a `run_id` that matches the scenario run. See the post-processing suggestion on the parent feature for this issue.

PR #253, opened 2025-08-19
feat: Add support for importing Holochain metrics into InfluxDB

ddd-mtl:
### Summary

https://github.com/holochain/wind-tunnel/issues/204

Note: For this to work, metrics to file support must be part of latest holochain 0.5.x version used by holonix

### TODO:

- [ ] All code changes are reflected in docs, including module-level docs

_Note that all commits in a PR must follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0) before it can be merged, as these are used to generate the changelog_

## Summary by CodeRabbit

- New Features
  - Optional Holochain metrics import flow to InfluxDB with run tagging.
  - Configurable run summary path via environment variable.
  - More descriptive run identifiers in CI and Nomad.
  - Improved Nomad job diagnostics with automatic failed-task log collection.
  - Host metrics collection/upload when using file-based reporters.

- Documentation
  - Added “Import Holochain Metrics” guide with examples and cleanup notes.
  - Fixed typos and clarified wording in metrics sections.

- Bug Fixes
  - Corrected malformed URL in README.

- Chores
  - Standardized Telegraf configs to use environment-driven RUN_ID and static paths.

PR #254, opened 2025-08-21
feat: Support for importing Holochain metrics into InfluxDB

ddd-mtl:
### Summary

Closes https://github.com/holochain/wind-tunnel/issues/204
For more context, see MM discussion: https://chat.holochain.org/holo/pl/gq54sjxktfdk3csuh9d891xpmw

Streamlined documentation on how to run scenarios locally with metrics. 
Some minor improvements to `import_lp_metrics` script and `lp-tool` for a smoother DX.
Removed or renamed unused telegraf related functions and files.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Centralized metrics importer that tags metrics with run_id, imports into InfluxDB, and cleans local metric files.
  * Option to enable Holochain metrics via a new environment variable.

* **Documentation**
  * Added an "Importing Metrics" section with run_id guidance, warnings, Nix-based workflow, and example commands.
  * Updated InfluxDB and host-metrics setup instructions.

* **Chores**
  * CI and packaging now invoke the new importer; Telegraf-specific configs and one-shot paths removed.

### DIFF
--- a/scripts/influx.sh
+++ b/scripts/influx.sh
@@ -41,7 +41,7 @@ clear_influx() {
 # Adds a run_id tag all metrics in $WT_METRICS_DIR and imports them into InfluxDB
 import_lp_metrics() {
     if [ -z "${INFLUX_TOKEN:-}" ]; then
-      echo "Environment variable INFLUX_TOKEN has not been set. Run 'use_influx' to set it." >&2
+        echo "Environment variable INFLUX_TOKEN has not been set. Run 'use_influx' to set it." >&2
         return 1
     fi
 


scripts/influx.sh
```diff
@@ -37,3 +37,44 @@ clear_influx() {
...
+          summary_path="$RUN_SUMMARY_PATH"
+      fi
+      if [ -f "$summary_path" ]; then
+          echo "Reading RUN_ID from file: $summary_path"
+          tmp_run_id=$(jq --slurp --raw-output 'sort_by(.started_at|tonumber) | last | .run_id' < "$summary_path")
+      else
+          echo "Run summary file not found: $summary_path" >&2
+          tmp_run_id=""
+      fi
+    fi
```
ddd-mtl, line 65: we want to use RUN_ID if it has been set, but we don't want to set it ourself here, so use a local var.

scripts/influx.sh
```diff
@@ -37,3 +37,44 @@ clear_influx() {
...
+    echo "Importing Holochain metrics from '$HOLOCHAIN_INFLUX_FILE' and adding tag run_id=$tmp_run_id"
+    # Create new file with added run_id tag at end of tag list of each metric
+    OUTPUT_FILE=$HOLOCHAIN_INFLUX_FILE.tmp.influx
+    sed "s/\([^[:space:]]*\) \(.*\)/\1,run_id=$tmp_run_id \2/" "$HOLOCHAIN_INFLUX_FILE" > "$OUTPUT_FILE"
+    # Import to influxDB
+    use_influx
```
ThetaSinner, line 71: Configures dev credentials, will not work on Nomad.

ThetaSinner (changes requested):
https://github.com/holochain/wind-tunnel/pull/253#discussion_r2290595350

- This PR doesn't address my previous concerns that the new Bash function overlaps with what the existing script does. If the existing flake script is run first then it will upload the metrics itself and this script will do nothing. I'm not clear why we need a separate mechanism for host metrics at all.
- Using `influx write` vs `telegraf --once` appears to me to be a trade-off. The `telegraf` tool has processor plugins which means if can add the extra `run_id` field to the output, where the `influx` tool requires modifying the raw metrics files which could easily go wrong with a change to the format or something we didn't know about the format of the run_id not respecting the required format.
- "This is simpler to use than telegraf and influx handles huge files better.", this needs more justification than just a statement if we're going to switch tooling. Is it faster? Does `telegraf` fail to process metrics in a batch when the file gets too big for its buffer and fail? Is the failure with `telegraf` a configuration issue with Telegraf or is it a genuine limitation of the tool? One way or another, if we're making the switch to another tool then we should actually make the switch and not have just host metrics doing things their own way. The existing scripts for Nomad and local metrics upload would need to be modified. Nothing that we would then be modifying more metrics files with `sed` which I think increases the need to carefully justify the switch from `telegraf --once` to `influx write`.

scripts/influx.sh
ThetaSinner: The issue is only about getting this working locally and not about Nomad. I still think that it's worth separating this out. It's part of the normal way of working with Influx in a dev environment for Wind Tunnel that this function needs to be run to set up your shell before running other commands.

scripts/influx.sh
```diff
@@ -37,3 +37,47 @@ clear_influx() {
...
+
+# Import Holochain metrics from $HOLOCHAIN_INFLUX_FILE into InfluxDB with added RUN_ID tag
+import_hc_metrics_into_influx() {
+    if [[ -z "${INFLUX_BUCKET:-}" ]] || [[ -z "${INFLUX_TOKEN:-}" ]]; then
+      echo "Environment variables INFLUX_BUCKET and INFLUX_TOKEN have not been set. Run 'use_influx' to have them set." >&2
+      return
+    fi
+    if [[ -z "$HOLOCHAIN_INFLUX_FILE" ]]; then
+        echo "HOLOCHAIN_INFLUX_FILE variable is not set. Skipping import of Holochain metrics."
+        return
```
veeso, line 49:
```suggestion
      return 1
```

scripts/influx.sh
```diff
@@ -37,3 +37,47 @@ clear_influx() {
...
+    fi
+    if [ ! -f "$HOLOCHAIN_INFLUX_FILE" ]; then
+        echo "HOLOCHAIN_INFLUX_FILE is set, but file is missing: $HOLOCHAIN_INFLUX_FILE. Make sure Holochain is running with this env variable."
+        return
```
veeso, line 53:
```suggestion
        return 1
```

scripts/influx.sh
```diff
@@ -37,3 +37,47 @@ clear_influx() {
...
+    fi
+    # Determine RUN_ID
+    local tmp_run_id=$RUN_ID
+    if [ -z "$tmp_run_id" ]; then
+      summary_path="run_summary.jsonl"
```
veeso, line 58:
```suggestion
      local summary_path="run_summary.jsonl"
```

scripts/influx.sh
```diff
@@ -37,3 +37,47 @@ clear_influx() {
...
+          echo "Reading RUN_ID from file: $summary_path"
+          tmp_run_id=$(jq --slurp --raw-output 'sort_by(.started_at|tonumber) | last | .run_id' < "$summary_path")
+      else
+          echo "Run summary file not found: $summary_path" >&2
+          tmp_run_id=""
+      fi
+    fi
+    echo "Importing Holochain metrics from '$HOLOCHAIN_INFLUX_FILE' and adding tag run_id=$tmp_run_id"
+    # Create new file with added run_id tag at end of tag list of each metric
+    OUTPUT_FILE=$HOLOCHAIN_INFLUX_FILE.tmp.influx
```
veeso, line 72: Why don't we use telegraf to import metrics here?

ddd-mtl: telegraf might overflow since the number of metrics in the file can exceed its buffer.

flake.nix
```diff
@@ -119,6 +127,7 @@
         packages = {
           default = config.workspace.workspace;
           inherit (config.workspace) workspace;
+          inherit lp-tool;
```
ddd-mtl, line 130: Make `lp-tool` available in nix shell

scripts/influx.sh
```diff
@@ -37,3 +37,47 @@ clear_influx() {
...
+    fi
+    echo "Importing Holochain metrics from '$HOLOCHAIN_INFLUX_FILE' and adding tag run_id=$tmp_run_id"
+    # Create new influx file with added run_id tag for each metric
+    local OUTPUT_FILE=$HOLOCHAIN_INFLUX_FILE-$tmp_run_id.influx
+    lp-tool -input "$HOLOCHAIN_INFLUX_FILE" -output "$OUTPUT_FILE" -tag run_id="$tmp_run_id"
+    # Import to influxDB
+    influx write \
+    --bucket "$INFLUX_BUCKET" \
+    --org holo \
+    --file "$OUTPUT_FILE"
```
ddd-mtl, line 80: `influx write` uses  $INFLUX_TOKEN and $INFLUX_HOST as default, so no need to specify these options.

ddd-mtl:
@ThetaSinner Replaced `sed` with `lp-tool`. Added `lp-tool` to nix shell.

scripts/influx.sh
```diff
@@ -37,3 +37,47 @@ clear_influx() {
...
+          echo "Reading RUN_ID from file: $summary_path"
+          tmp_run_id=$(jq --slurp --raw-output 'sort_by(.started_at|tonumber) | last | .run_id' < "$summary_path")
+      else
+          echo "Aborting. RUN_ID not set and run summary file not found: $summary_path" >&2
+          return 1
+      fi
+    fi
+    echo "Importing Holochain metrics from '$HOLOCHAIN_INFLUX_FILE' and adding tag run_id=$tmp_run_id"
+    # Create new influx file with added run_id tag for each metric
+    local OUTPUT_FILE=$HOLOCHAIN_INFLUX_FILE-$tmp_run_id.influx
```
veeso, line 74: use mktemp to create a temp file

veeso (changes requested):

scripts/influx.sh
```diff
@@ -37,3 +37,49 @@ clear_influx() {
...
+          tmp_run_id=$(jq --slurp --raw-output 'sort_by(.started_at|tonumber) | last | .run_id' < "$summary_path")
+      else
+          echo "Aborting. RUN_ID not set and run summary file not found: $summary_path" >&2
+          return 1
+      fi
+    fi
+    echo "Importing Holochain metrics from '$HOLOCHAIN_INFLUX_FILE' and adding tag run_id=$tmp_run_id"
+    # Create new influx file with added run_id tag for each metric
+    local tmp_output_file
+    tmp_output_file=$(mktemp -u)
```
ddd-mtl, line 75: `-u` option because `lp-tool` will overwrite the file and print a warning if it already exists, so we just need a dry run to get the temp file name

scripts/influx.sh
```diff
@@ -37,3 +37,49 @@ clear_influx() {
...
+    lp-tool -input "$HOLOCHAIN_INFLUX_FILE" -output "$tmp_output_file" -tag run_id="$tmp_run_id"
+    # Import to influxDB
+    influx write \
+    --bucket "$INFLUX_BUCKET" \
+    --org holo \
+    --file "$tmp_output_file"
+    # Clean up temp file
+    rm -f "$tmp_output_file"
+    echo "Successfully imported Holochain metrics to '$INFLUX_HOST' and deleted file $tmp_output_file"
```
ddd-mtl, line 84: `lp-tool` will print a message saying the tmp file has been created, but it will deleted after soon after, so notify the user that this file has already been deleted.

veeso (approved):

.github/workflows/performance.yaml
```diff
@@ -54,22 +54,9 @@ jobs:
...
-            grep "E!" logs/telegraf-stderr.log | xargs -l echo "::error ::"
-            echo "::endgroup::"
-          fi
-
-          echo "::group::Telegraf warnings"
-          # Print warnings as such in GitHub logs.
-          grep "W!" logs/telegraf-stderr.log | xargs -l echo "::warning ::"
-          echo "::endgroup::"
-
+          run .#ci-import-metrics
```
ddd-mtl, line 59: Renamed #ci-telegraf to #ci-import-metrics since its not using telegraf anymore.
There are no telegraf logs to grep anymore.

README.md
```diff
@@ -580,45 +617,6 @@ RUST_LOG=info cargo run -p kitsune_continuous_flow -- --bootstrap-server-url htt
...
-nix run .#local-telegraf
-```
-
-At this point the metrics will be imported to InfluxDB and you will be able to view the Host metrics in the InfluxDB Host dashboard by run_id.
-
-Running the `telegraf` agent will also clean up the previous metrics, so you are immediately ready to run the next scenario.
-
-> [!Warning]
-> The metrics must be imported after each scenario run and cleanup, since they are associated only to the latest scenario run.
-
```
ddd-mtl, line 621: This has been moved within the `Running scenarios locally` section.

flake.nix
```diff
@@ -150,15 +149,14 @@
...
-            name = "ci-telegraf";
+          ci-import-metrics = pkgs.writeShellApplication {
+            name = "ci-import-metrics";
             runtimeInputs = [
               lp-tool
               pkgs.gnused
               pkgs.influxdb2-cli
               pkgs.jq
               pkgs.yq
-              unstablePkgs.telegraf
```
ddd-mtl, line 161: telegraf not used anymore here.

lp-tool/main.go
```diff
@@ -181,7 +181,7 @@ func processLineProtocol(config Config) error {
 			// Add custom tags
             for key, value := range config.Tags {
                 // Warn if tag already exists
-                if tag, has := telegrafMetric.GetTag(key) ; has {
+                if tag, has := telegrafMetric.GetTag(key) ; has  && tag != value {
```
ddd-mtl, line 184: Don't print a warning if the run_id is overwritten with the same value

telegraf/telegraf.conf
```diff
@@ -1,176 +0,0 @@
...
-[[inputs.tail]]
-    files = ["$WT_METRICS_DIR/*.influx"]
-    from_beginning = false
-    pipe = false
-    watch_method = "inotify"
-    max_undelivered_lines = 1000
-    character_encoding = "utf-8"
-    data_format = "influx"
-    path_tag = ""
-    filters = []
```
ddd-mtl, line 176: I didn't see this file get used anywhere, so deleting.

telegraf/telegraf.once.conf
```diff
@@ -1,34 +0,0 @@
...
-  bucket = "windtunnel"
-
-###############################################################################
-#                            INPUT PLUGINS                                    #
-###############################################################################
-
-[[inputs.file]]
-  files = ["${WT_METRICS_DIR}/*.influx"]
-  character_encoding = "utf-8"
-  data_format = "influx"
```
ddd-mtl, line 34: Again. I didn't see this file get used anywhere, so deleting.

.github/workflows/performance.yaml
```diff
@@ -54,22 +54,9 @@ jobs:
...
-            grep "E!" logs/telegraf-stderr.log | xargs -l echo "::error ::"
-            echo "::endgroup::"
-          fi
-
-          echo "::group::Telegraf warnings"
-          # Print warnings as such in GitHub logs.
-          grep "W!" logs/telegraf-stderr.log | xargs -l echo "::warning ::"
-          echo "::endgroup::"
-
+          nix run .#ci-import-metrics
```
ddd-mtl, line 59: Renamed #ci-telegraf to #ci-import-metrics since its not using telegraf anymore.
Also, there are no telegraf logs to grep anymore.

scripts/influx.sh
```diff
@@ -38,9 +38,10 @@ clear_influx() {
     rm "$INFLUX_CONFIGS_PATH"
 }
 
+# Adds a run_id tag all metrics in $WT_METRICS_DIR and imports them into InfluxDB
 import_lp_metrics() {
     if [ -z "${INFLUX_TOKEN:-}" ]; then
-        echo "INFLUX_TOKEN is not set"
+      echo "Environment variable INFLUX_TOKEN has not been set. Run 'use_influx' to set it." >&2
```
veeso, line 44:
```suggestion
        echo "Environment variable INFLUX_TOKEN has not been set. Run 'use_influx' to set it." >&2
```

veeso (commented):
It's good, apart from a little indentation typo on the script.

I'm not sure whether we want to have these changes in a single PR or have one PR for the docs for holochain metrics and another to remove unused telegraf names/conf/scripts etc

So I want to wait for other's opinions before approving

README.md
```diff
@@ -369,9 +389,26 @@ Each scenario is expected to provide a README.md with at least:
 
 For example, see the [zome_call_single_value](https://github.com/holochain/wind-tunnel/blob/main/scenarios/zome_call_single_value/README.md) scenario.
 
-As well as the command you use to run the scenario, you will need to select an appropriate reporter. Try running the scenario with the `--help` flag to see the available options.
-For local development, the default `in-memory` reporter will do. If you have influx running then you can use the `influx-client` option. If you have both Influx and Telegraf running
-then you can use the `influx-file` option.
```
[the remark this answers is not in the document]
ThetaSinner: I don't agree with removing these lines entirely. The `in-memory` reporter is the default. The `influx-client` option should probably be removed from the code. Then the `influx-file` option is the one you need when InfluxDB is running. The documentation just needs updating because it *was* uploaded with Telegraf previously and it was done in real time as the scenario ran. Instead, a reference to the local import script would be useful here instead of `telegraf`.
 
 Also worth noting that the documentation on the `InfluxFile` enum variant is out of date.

README.md
```diff
@@ -369,9 +386,25 @@ Each scenario is expected to provide a README.md with at least:
...
+
+Once you've finished running a scenario, you can collect host, Holochain and scenario metrics with
+
+```sh
+nix run .#local-import-metrics
+```
+
+At this point the metrics will be imported to InfluxDB, and you will be able to view the metrics in the InfluxDB dashboards by run_id.
+
+Running this nix command will also clean up the current metrics from disk, so you are immediately ready to run the next scenario.
```
ThetaSinner, line 402:
```suggestion
Running this Nix command will also clean up the current metrics from disk, so you are immediately ready to run the next scenario.
```

README.md
```diff
@@ -369,9 +386,25 @@ Each scenario is expected to provide a README.md with at least:
...
+
+> [!Warning]
+> The metrics must be imported after each scenario run since they are associated only with the latest scenario run.
+> [!Warning]
+> If Holochain ran with metrics enabled, it must be restarted after each scenario run since its output file is deleted after importing.
```
ThetaSinner, line 407: Isn't the same true of host metrics? That should be restarted too?

veeso (approved):

Issue #109, opened 2024-10-15
Collect metrics from target nodes


ThetaSinner:
When we run TryCP-based scenarios, we are running one control node and connecting to several targets that have TryCP and Holochain on them. We collect custom and host metrics from the control node but we don't yet collect any metrics from Holochain or the host on the target nodes.

We could solve this in two ways
- Modify Holochain to integrate with Telegraf, and have TryCP launch Holochain with that enabled. At the end of the scenario we could have TryCP run Telegraf to push metrics to InfluxDB
- Investigate a tool that could run Wind Tunnel scenarios the way that they were originally designed to run. That is, without TryCP. Then figure out how to integrate Telegraf with that tool.

ThetaSinner:
I suggest that we do #111 before this, then we can decide which option we're going with and can add some AC for this issue.

PR #256, opened 2025-08-26
feat: New CLI tool `lp-tool` for processing InfluxDB line protocol files

ddd-mtl:
## Summary

As part of https://github.com/holochain/wind-tunnel/issues/109 for adding the `run_id` to metrics produced by holochain.

The CLI tool reads a line protocol input file. Converts each valid line into an internal representation of a telegraf metric. Adds a `run_id` tag  to each metric with a value provided via a CLI option, then serialize the modified metric to a new line protocol output file.

For more context see MM thread: https://chat.holochain.org/holo/pl/gq54sjxktfdk3csuh9d891xpmw
and "downstream" PR: https://github.com/holochain/wind-tunnel/pull/254

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added a Line Protocol CLI to append custom tags to InfluxDB line protocol files (configurable input/output and repeatable tag flags).

* **Documentation**
  * Included a usage guide with install/build steps, options, examples, and dependency notes for the CLI.

* **Tests**
  * Added comprehensive tests for tag parsing, CLI flags, processing behavior, and I/O error handling.

* **Chores**
  * Enabled Go tooling in local/CI dev environments and added Go checks to CI and project scripts.


lp-tool/main.go
```diff
@@ -0,0 +1,141 @@
...
+
+import (
+	"bufio"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/influxdata/telegraf/plugins/parsers/influx"
+	influxSer "github.com/influxdata/telegraf/plugins/serializers/influx"
```
ddd-mtl, line 11: need an alias as it would conflict with the other telegraf dep

lp-tool/main_test.go
```diff
@@ -0,0 +1,324 @@
+package main
```
ddd-mtl, line 1: generated tests

scripts/checks.sh
```diff
@@ -20,10 +20,18 @@ check_rust_static() {
   cargo clippy --workspace --all-targets --all-features -- --deny warnings
 }
 
+check_go() {
+  go mod tidy
+  go build -C lp-tool -o lp-tool
+  go test -v ./...
+  ./lp-tool/lp-tool -h
```
ddd-mtl, line 27: script called by CI to check if go app is functional

cdunster (commented):
Maybe I'm missing some context because I'm a bit confused as to why this should be added? Why is it written in Go? Is it because Telegraf has some Go code that can be imported to do this? I don't have an issue with Go, if it makes sense, but I'm not sure why the choice was made.

Also, you mentioned that this is to replace the use of sed, but then I don't understand why we would support a whole application, written in a language that we typically don't use, to replace a widely-used and well-maintained Unix utility?

lp-tool/README.md
```diff
@@ -0,0 +1,48 @@
...
+```bash
+./lp-tool [options]
+```
+
+### Options
+
+- `-input`: Input line protocol file path (default: "holochain.influx")
+- `-output`: Output line protocol file path (default: "holochain.tmp.influx") 
+- `-tag-name`: Name of the custom tag to add (default: "run_id")
+- `-tag-value`: Value of the custom tag to add (required)
```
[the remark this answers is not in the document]
veeso: Mhm, not a fan of this, maybe I would use an arg `-tag <tagname>=<value>` which can be used multiple times.

lp-tool/README.md
```diff
@@ -0,0 +1,48 @@
+# Line Protocol CLI Tool
+
+A command-line tool that reads InfluxDB line protocol files, converts them to Telegraf metrics, adds a custom tag, and writes the processed data to a new line protocol file.
```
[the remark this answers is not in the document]
ddd-mtl: I'm referring to the internal code representation used by telegraf: https://pkg.go.dev/github.com/influxdata/telegraf@v1.35.4#Metric

Being text, line protocol is more of a serialization format.

lp-tool/main.go
```diff
@@ -0,0 +1,141 @@
...
+	processedCount := 0
+
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		lineCount++
+
+		// Skip empty lines and comments
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
```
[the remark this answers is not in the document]
ddd-mtl: it is a possibility if the input files are edited manually for whatever reason.

lp-tool/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```bash
+go mod tidy
+```
+
+3. Build the executable:
+
+```bash
+go build -C lp-tool -o lp-tool
+```
```
cdunster, line 19: Currently, I think that the ordered lists and code blocks are actually separate markdown blocks; you need to indent by 4 spaces to make the code blocks part of the list.

```suggestion

    ```bash
    go mod tidy
    ```

3. Build the executable:

    ```bash
    go build -C lp-tool -o lp-tool
    ```
```

lp-tool/main.go
```diff
@@ -0,0 +1,189 @@
...
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "\nExample:\n")
+		fmt.Fprintf(os.Stderr, "  %s -input data.influx -output holochain.influx -tag env=production -tag region=europe -tag version=1.2.3\n", os.Args[0])
+	}
+
+	flag.Parse()
+
+	if config.InputFile == "" || len(config.Tags) == 0 {
+		return config, fmt.Errorf("input file or tags not specified")
+	}
```
cdunster, line 96: I think these would be better as separate checks with different logs. Also, the input file has a default so an empty string would mean that someone specified `-input=""`

```suggestion
	if config.InputFile == "" {
		return config, fmt.Errorf("input file cannont be an empty path")
	}
		if len(config.Tags) == 0 {
		return config, fmt.Errorf("No tags were specified")
	}
```

lp-tool/main.go
```diff
@@ -0,0 +1,189 @@
...
+    if config.OutputFile == "" {
+        config.OutputFile = config.InputFile + ".tmp.influx"
+    }
```
cdunster, line 99: Shouldn't this also just be an error? It seems strange that someone could specify `-output=""` and they get a seemingly random output file.

```suggestion
    if config.OutputFile == "" {
		return config, fmt.Errorf("output file cannont be an empty path")
    }
```

lp-tool/main.go
```diff
@@ -0,0 +1,189 @@
...
+func processLineProtocol(config Config) error {
+	// Open input file
+	inputFile, err := os.Open(config.InputFile)
+	if err != nil {
+		return fmt.Errorf("failed to open input file: %w", err)
+	}
+	defer inputFile.Close()
+
+	// Create output file
+	outputFile, err := os.Create(config.OutputFile)
```
cdunster, line 115: It might be nice if we prompt to overwrite an existing file first? Maybe add a flag to force it for scripts.

ddd-mtl: How about keeping it simple and just printing a warning that the file will be overwritten?

ddd-mtl:
Changed the api the have input required.
Better handling of invalid options.
If output is unset, use input and change extension to `.tmp.influx` 
Warn when output file will be overwritten.

lp-tool/main.go
cdunster: Sure. I don't think that helps much, though, because it's still a destructive and irreversible action, and the user won't see the warning until it's too late. I'm fine to leave this as is because of how the tool will be used; it's just potentially dangerous.

lp-tool/main.go
```diff
@@ -91,15 +94,28 @@ func parseFlags() (Config, error) {
 
 	flag.Parse()
 
-	if config.InputFile == "" || len(config.Tags) == 0 {
-		return config, fmt.Errorf("input file or tags not specified")
+    // Check for invalid options.
+	if config.InputFile == "" {
+		return config, fmt.Errorf("empty input file specified")
```
cdunster, line 99: Nitpick, so only change if you're making other changes:
This reads to me that the file itself is empty, even though what you have is technically correct, I would suggest something like:

```suggestion
		return config, fmt.Errorf("no input file specified")
```

lp-tool/main.go
```diff
@@ -91,15 +94,28 @@ func parseFlags() (Config, error) {
...
+	}
+	if len(config.Tags) == 0 {
+		return config, fmt.Errorf("tags not specified")
 	}
     if config.OutputFile == "" {
-        config.OutputFile = config.InputFile + ".tmp.influx"
+        return config, fmt.Errorf("empty output file specified")
```
cdunster, line 105: Same nitpick here:

```suggestion
        return config, fmt.Errorf("no output file specified")
```

cdunster (dismissed):
Some nitpicks and suggestions, but I'm happy with the current state, so feel free to ignore the suggestions.

veeso (dismissed):

lp-tool/README.md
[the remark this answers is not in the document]
ddd-mtl: I removed the reference to Telegraf Metrics from the readme and added it to the PR description instead.

lp-tool/main.go
ddd-mtl: will fix

lp-tool/main.go
ddd-mtl: I think it still helps as the user can spot a potential issue with their workflow, and can also use whatever backup mechanism they have to retrieve the file if needed. As you say, given the context I believe this is good enough.

lp-tool/main.go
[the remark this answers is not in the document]
ThetaSinner: I don't understand how that response answers my question.

I'm asking whether the `#` character is valid "start line comment" syntax in the Influx line protocol.

These files are machine generated and should never be edited by hand.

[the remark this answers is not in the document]
ddd-mtl: Yes starting `#` is part of the syntax: https://docs.influxdata.com/influxdb/v2/reference/syntax/line-protocol/#comments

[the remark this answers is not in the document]
ThetaSinner: Great, okay. In that case fair enough with this filter.

jost-s:
Tests look fine to me. I've ran the tool successfully. There's just the question about the output file. If we keep the input file unchanged, then I prefer another file extension than `.tmp*` for the output file.

lp-tool/README.md
```diff
@@ -0,0 +1,48 @@
...
+2. Initialize the Go module and download dependencies:
+
+    ```bash
+    go mod tidy
+    ```
+
+3. Build the executable:
+
+    ```bash
+    go build -C lp-tool -o lp-tool
```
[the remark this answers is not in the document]
ddd-mtl: readme updated

go.mod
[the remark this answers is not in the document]
ddd-mtl: yes, I moved the go build files into the subfolder

ddd-mtl:
@ThetaSinner Moved Go build files for lp-tool to its subfolder. Updated CI and README.md accordingly and according to suggestions.

ThetaSinner (approved):
Looks good

PR #272, opened 2025-09-10
ci: parallelise the running of nomad jobs

### LINKED ISSUE #268: Parallelise the running of Nomad jobs
Currently, the Nomad CI workflow waits for each Nomad job to finish before that job is marked as completed, see:
https://github.com/holochain/wind-tunnel/blob/a124cd57ad7fc257b11e25ba25c336486e80e10b/.github/workflows/nomad.yaml#L114-L119

The job also has `max-parallel` set to `1`, see:
https://github.com/holochain/wind-tunnel/blob/a124cd57ad7fc257b11e25ba25c336486e80e10b/.github/workflows/nomad.yaml#L15

These things together mean that only a single scenario can be run at once. To speed up the running of all the scenarios, it would be better to only be limited by the number of online Nomad clients. I see two ways of doing this:

1. Just remove the setting of `max-parallel` so that all the CI jobs can run at once. This might cause problems with scenarios that require multiple agents, as each CI job might request a Nomad job to be started on the same Nomad client at the same time, which will cause unexpected and undefined behaviour if we don't have enough clients, because of the way that Nomad works. It will just wait for a client to be available, but will still start the job on the other clients used by the scenario. E.g., if a scenario required two agents and only one is available, then Nomad will start on the one available client, causing the scenario to fail as it couldn't communicate with the other (non-existent) agent.
2. Collect a list of started Nomad allocations and then wait for them all at the end of the CI workflow. This means that each Nomad job is started one after another, and we can wait for enough clients to be available before doing so, but once it's running, then we start the next one. Then, only once all scenarios are running do we wait for them to finish.

veeso:
### Summary

closes #268

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Waits on multiple Nomad allocations across all scenarios with a dedicated aggregate wait step.
  * Per-run allocation IDs are saved as artifacts and aggregated for cross-run monitoring.
  * Added a small pause when waiting for free nodes to improve resource availability handling.

* **Bug Fixes / Reliability**
  * Enforced presence of allocation IDs before proceeding, improving error reporting.

* **Chores**
  * Added automatic restart policy and increased CPU for the Holochain task.

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -75,6 +75,7 @@ jobs:
               else
                   echo -n "."
               fi
+              sleep 5
           done
           echo "done"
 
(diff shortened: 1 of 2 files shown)


nomad/run_scenario.tpl.hcl
```diff
@@ -77,6 +77,12 @@ job "{{ (ds "vars").scenario_name }}" {
     labels   = ["${var.scenario-name}-${group.key}-${group.value}"]
 
     content {
+      restart {
+        interval         = "30m"
```
jost-s, line 81: What does the interval define?

.github/workflows/nomad.yaml
```diff
@@ -103,17 +103,46 @@ jobs:
...
           NOMAD_CACERT: "${{ github.workspace }}/nomad/server-ca-cert.pem"
           NOMAD_TOKEN: ${{ secrets.NOMAD_ACCESS_TOKEN }}
-        run: nix develop --command ./nomad/wait_for_jobs.sh ${{ matrix.job-name || matrix.scenario-name }} ${{ steps.run-nomad-job.outputs.alloc_id }}
+        run: |-
+          cat ./alloc_ids/*/alloc_ids.csv > all_alloc_ids.csv
+          while IFS= read -r line; do
+            scenario_name=$(echo "$line" | cut -d',' -f1)
+            alloc_id=$(echo "$line" | cut -d',' -f2)
+            echo "Waiting for $scenario_name with allocation ID $alloc_id"
+            nix develop --command ./nomad/wait_for_jobs.sh $scenario_name $alloc_id
```
jost-s, line 147: Should there be a sleep between retries in here or in the script?

nomad/run_scenario.tpl.hcl
veeso: https://developer.hashicorp.com/nomad/docs/job-specification/restart

[interval](https://developer.hashicorp.com/nomad/docs/job-specification/restart#interval) (string: <varies>) - Specifies the duration which begins when the first task starts and ensures that only attempts number of restarts happens within it. If more than attempts number of failures happen, behavior is controlled by mode. This is specified using a label suffix like "30s" or "1h". Defaults vary by job type, see below for more information.

jost-s: Thanks, so in this combination it limits the overall duration of a job, right?

veeso: @jost-s Exactly

PR #274, opened 2025-09-15
chore(nomad): Cleanup Nomad job spec variables

### LINKED ISSUE #267: Cleanup Nomad job-spec variables
The [Nomad job-spec template file](https://github.com/holochain/wind-tunnel/blob/a124cd57ad7fc257b11e25ba25c336486e80e10b/nomad/run_scenario.tpl.hcl) has a list of variables at the top. However, some of them are no longer needed and thus can be removed, some have defaults that shouldn't, the templating currently inserts a blank line between the description and the default values, and they should be renamed to use underscores so that they work better with setting them via environment variables: `NOMAD_VAR_some_var` as opposed to `NOMAD_VAR_some-var`.

AC:
- [ ] Remove unnecessary variables that won't change between runs, i.e. `scenario-name`.
- [ ] Remove undesired defaults, i.e. `scenario-url`.
- [ ] Update the `gomplate` comment about the defaults to remove the blank line in the generated file.
- [ ] Rename all variables to use underscores instead of hyphens.
- [ ] Update the Nomad CI workflow to use the new variable names

veeso:
# Summary

- Removed unnecessary variables that won't change between runs, i.e. scenario-name.
- Remove undesired defaults
- Update the gomplate comment about the defaults to remove the blank line in the generated file
- Rename all variables to use underscores instead of hyphens
- Update the Nomad CI workflow to use the new variable names
- Update README based on changes, i.e., no scenario name and different var names

closes #267

### TODO:

## Summary by CodeRabbit

- Refactor
  - Simplified Nomad job configuration, standardized parameter names to snake_case, centralized settings into data-driven defaults, and consolidated run identifiers.

- Documentation
  - Updated local Nomad instructions: examples now use snake_case parameters, add an explicit step to generate Nomad jobs before running, and adjusted run command examples accordingly.

- Chores
  - CI workflow streamlined by removing the download step and aligning environment variables and run-id handling with the new configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -562,6 +562,12 @@ accessible.
 > Unlike when running locally in the section above, we cannot just pass a path because the path needs to be
 > accessible to the client and Nomad doesn't have native support for uploading artefacts.
 
+At this point you have to generate the Nomad job for the scenario you want to run. This is done with:
+
+```bash
+./nomad/generate_jobs.sh app_install_minimal
+```
+
 Now that the bundle is publicly available you can run the scenario with the following:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -477,24 +477,24 @@ Replace `app_install` with the name of the scenario that you want to run.
 Once the scenario is built you can run the Nomad job with:
 
 ```bash
-nomad job run -address=http://localhost:4646 -var scenario-url=result/bin/app_install -var reporter=in-memory nomad/jobs/app_install_minimal.nomad.hcl
+nomad job run -address=http://localhost:4646 -var scenario_url=result/bin/app_install -var reporter=in-memory nomad/jobs/app_install_minimal.nomad.hcl
 ```
 
 All the jobs are in the `nomad/jobs` directory, so you can replace `app_install_minimal` with the name of the job you want to run.
 
 - `-address` sets Nomad to talk to the locally running instance and not the dedicated Wind Tunnel cluster one.
-- `-var scenario-url=...` provides the path to the scenario binary that you built in the previous step.
+- `-var scenario_url=...` provides the path to the scenario binary that you built in the previous step.
 - `-var reporter=in-memory` sets the reporter type to print to `stdout` instead of writing an InfluxDB metrics file.
 
 > [!Warning]
-> When running locally as in this guide, the `reporter` must be set to `in-memory` and the `scenario-url` must be a
+> When running locally as in this guide, the `reporter` must be set to `in-memory` and the `scenario_url` must be a
 > local path due to the way Nomad handles downloading. To get around this limitation you must disable file system
 > isolation, see <https://developer.hashicorp.com/nomad/docs/configuration/client#disable_filesystem_isolation>.
 
 You can also override existing and omitted variables with the `-var` flag. For example, to set the duration (in seconds) use:
 
 ```bash
-nomad job run -address=http://localhost:4646 -var scenario-url=result/bin/app_install -var reporter=in-memory -var duration=300 nomad/jobs/app_install_minimal.nomad.hcl
+nomad job run -address=http://localhost:4646 -var scenario_url=result/bin/app_install -var reporter=in-memory -var duration=300 nomad/jobs/app_install_minimal.nomad.hcl
 ```
 
 > [!Note]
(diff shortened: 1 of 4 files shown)


nomad/run_scenario.tpl.hcl
```diff
@@ -1,65 +1,36 @@
...
-  description = "The name of the scenario to run"
-  default = {{ (ds "vars").scenario_name | quote }}
-}
-
-variable "connection-string" {
+variable "connection_string" {
   type        = string
   description = "The URL to be used to connect to the service being tested"
-  {{/* Default: read `connection_string` from the JSON data source `vars`, or set to `"ws://localhost:8888"` if not provided.*/}}
+  {{- /* Default: read `connection_string` from the JSON data source `vars`, or set to `"ws://localhost:8888"` if not provided.*/}}
```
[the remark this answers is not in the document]
veeso: @jost-s 

Removes the empty line, which was the purpose.

So this

```txt
variable "connection-string" {
  type        = string
  description = "The URL to be used to connect to the service being tested"
  {{/* Default: read `connection_string` from the JSON data source `vars`, or set to `"ws://localhost:8888"` if not provided.*/}}
  default     = {{ index (ds "vars") "connection_string" | default "ws://localhost:8888" | quote }}
}
```

Caused this

```txt
variable "connection_string" {
  type        = string
  description = "The URL to be used to connect to the service being tested"

  default     = "ws://localhost:8888"
}
```

With `{{ /`

it removes the empty line

```txt
variable "connection_string" {
  type        = string
  description = "The URL to be used to connect to the service being tested"
  default     = "ws://localhost:8888"
}
```

jost-s:
@veeso Please explain

> Why did you move the agents-per-node and min-agents directly to the config?

veeso:
@jost-s Sorry, I missed your comment there. I think the value depends only on the scenario, so when the scenario is generated with the config, the value should already be correct, and I don't think it makes sense to change it after that. But maybe @cdunster can confirm whether it should be configurable or not.

cdunster:
> @jost-s Sorry, I missed your comment there. I think the value depends only on the scenario, so when the scenario is generated with the config, the value should already be correct, and I don't think it makes sense to change it after that. But maybe @cdunster can confirm whether it should be configurable or not.

Yeah, I imagine that we want to keep the runs the same as much as possible, so I'm happy with this change.

`min-agents` should always be `behaviours.len() * agents_per_node`, I'm not sure if that's possible in `gomplate`?

nomad/run_scenario.tpl.hcl
```diff
@@ -152,32 +123,32 @@ job "{{ (ds "vars").scenario_name }}" {
...
         dynamic "artifact" {
           // Download the scenario from the URL if it is not a valid local path.
-          for_each = fileexists(abspath(var.scenario-url)) ? [] : [var.scenario-url]
+          for_each = fileexists(abspath(var.scenario_url)) ? [] : [var.scenario_url]
 
           content {
-            source = var.scenario-url
+            source = var.scenario_url
           }
         }
```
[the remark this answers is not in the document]
veeso: After discussing with Callum we've decided to keep as-is, because it wouldn't simplify the flow.

nomad/run_scenario.tpl.hcl
```diff
@@ -152,32 +123,32 @@ job "{{ (ds "vars").scenario_name }}" {
...
 
         env {
           RUST_LOG          = "info"
           HOME              = "${NOMAD_TASK_DIR}"
           WT_METRICS_DIR    = "${NOMAD_ALLOC_DIR}/data/telegraf/metrics"
-          MIN_AGENTS        = "${var.min-agents}"
+          MIN_AGENTS        = "{{ math.Mul (index (ds "vars") "agents_per_node" | default 1) 2 }}"
```
cdunster, line 137: This doesn't look right, it should be `agents_per_node * behaviours.len()` if possible. There's not always 2 behaviours.

This is because each behaviour gets run on a separate node, so 2 behaviours means 2 nodes. Then the `agents_per_node` allows us to run multiple agents on a single node, setting this to e.g. 5 means that each node runs 5 agents. So, if there are 2 behaviours, then there are 2 nodes, each with 5 agents, so the `MIN_AGENTS` should be `10`. But if there is only one behaviour, then `MIN_AGENTS` should be 5.

README.md
```diff
@@ -565,17 +565,17 @@ accessible.
 Now that the bundle is publicly available you can run the scenario with the following:
 
 ```bash
-nomad job run -var-file=nomad/var_files/app_install_minimal.vars -var scenario-url=http://{some-url} nomad/run_scenario.nomad.hcl
+nomad job run -var scenario_url=http://{some-url} nomad/jobs/app_install_minimal.nomad.hcl
```
cdunster, line 568: Should probably mention generating the job file before running this command.

cdunster (changes requested):

nomad/run_scenario.tpl.hcl
veeso: Yeah sorry, you're right.

PR #277, opened 2025-09-17
Feat/write get agent activity

mattyg:
### Summary
Scenario for measuring the time to get_agent_activity with an increasing chain length.

This PR relies on [fix/get_peer_list_randomized_remote_agents](https://github.com/holochain/wind-tunnel/tree/fix/get_peer_list_randomized_remote_agents). It is set to merge into that branch for a clean diff.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added a write_get_agent_activity scenario allowing peers to write entries and query another peer's activity; new coordinator and integrity zome components support this.

* **Observability**
  * New dashboard panels and reporting/summarization for the scenario (metrics, timing, and host summaries).

* **Tests**
  * CI workflows and a smoke test added to run the new scenario; test snapshots and run summaries included.

* **Documentation**
  * Scenario README with usage and run instructions.

* **Chores**
  * Workspace and pipeline updated to include new packages and configs.


scenarios/write_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+        ScenarioDefinitionBuilder::<
+            HolochainRunnerContext,
+            HolochainAgentContext<ScenarioValues>
+        >::new_with_init(
+            env!("CARGO_PKG_NAME"),
+        )
+        .with_default_duration_s(60)
+        .use_setup(setup)
+        .use_agent_setup(agent_setup)
+        .use_named_agent_behaviour("write", agent_behaviour_write)
```
mattyg, line 107: `write` behavior just creates entries, `get_agent_activity` behavior just gets the activity of the `write` peer.

scenarios/write_get_agent_activity/README.md
```diff
@@ -0,0 +1,15 @@
+## write_get_agent_activity
+
+### Description
+
+An agent creates an entry, then gets their own agent activity with `get_agent_activity`
```
mattyg, line 5:
```suggestion
An agent creates entries, while another agent gets their agent activity with `get_agent_activity`.
```

summariser/src/scenario/write_get_agent_activity.rs
```diff
@@ -0,0 +1,43 @@
...
+    assert_eq!(summary.scenario_name, "write_get_agent_activity");
+
+    let frame = query::query_custom_data(
+        client.clone(),
+        &summary,
+        "wt.custom.write_get_agent_activity",
+        &["agent"],
+    )
+    .await
+    .context("Load write_get_agent_activity data")?;
```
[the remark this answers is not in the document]
mattyg: fixed

scenarios/write_get_agent_activity/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+An agent creates entries, while another agent gets their agent activity with `get_agent_activity`.
+
+### Suggested command
+
+You can run the scenario locally with the following command:
+
+```bash
+RUST_LOG=info MIN_AGENTS=2 cargo run --package write_get_agent_activity -- --agents 2 --behaviour write:1 --behaviour get_agent_activity:1 --duration 60
+```
```
ThetaSinner, line 13: Possibly worth explicitly mentioning the interaction between sync and `get_agent_activity` here? If the two nodes reach sync with each other then the writer will publish their created data and `get_agent_activity` should often have most of the data it needs locally. So this is (correctly) more a test of database query performance and the code path through the host function than it is about `get_agent_activity` in general?

scenarios/write_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+                    .with_field(
+                        "highest_observed_action_seq",
+                        activity.highest_observed.map_or(0, |v| v.action_seq),
+                    )
+                    .with_field("value", elapsed.as_secs_f64()),
+            );
+        }
+        _ => {
+            if let Some(write_peer) = get_peer_list_randomized(ctx)?.first() {
+                ctx.get_mut().scenario_values.write_peer = Some(write_peer.clone());
```
ThetaSinner, line 78: This picks any remote peer on the network regardless of their behavior. This effectively means this scenario can only be used with two peers. Ideally scenarios are flexible enough that they can be run with any number of peers and any partition of behaviors across those peers. Consider the pattern in `two_party_countersigning` which deals with this problem by using DHT data to select peers rather than using `get_peer_list_randomized`?

scenarios/write_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+            let agent_pub_key = ctx.get().cell_id().agent_pubkey().to_string();
+            reporter.add_custom(
+                ReportMetric::new("write_get_agent_activity")
+                    .with_tag("agent", agent_pub_key)
+                    .with_field(
+                        "highest_observed_action_seq",
+                        activity.highest_observed.map_or(0, |v| v.action_seq),
+                    )
+                    .with_field("value", elapsed.as_secs_f64()),
+            );
```
ThetaSinner, line 74: Is it necessary to record a custom metric here? The `now` is just measuring across the `call_zome` function which covers a little more than just the `client.call_zome` but only response decoding and launching a task. I think it would be valid here to just use the automatically captured metric for the `client.call_zome` and filter that by function name.

summariser/src/scenario/write_get_agent_activity.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    let host_metrics = HostMetricsAggregator::new(&client, &summary)
+        .try_aggregate()
+        .await;
```
ThetaSinner, line 32: I'm surprised to not see a similar line for Holochain metrics here. That's not a problem for this PR but something to be looked at in general! :)

zomes/agent_activity/integrity/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+name = "agent_activity_integrity"
+path = "src/lib.rs"
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+hdi = { workspace = true }
+holochain_serialized_bytes = { workspace = true }
+serde = { workspace = true }
+
+# TODO watiing for a fix in 0.2.7-rc.1+
```
ThetaSinner, line 18: Typo in `waiting`. This won't be a problem for much longer, 0.6 of Holochain upgrades to 0.3 of this library.

scenarios/write_get_agent_activity/src/main.rs
mattyg: Fixed: https://github.com/holochain/wind-tunnel/pull/277/commits/5c1b95ae108a9946befd86873b8e0077f43c6278

Now the `get_agent_activity` peer picks a random write peer. Does that make sense?

scenarios/write_get_agent_activity/README.md
mattyg: Expended the description: https://github.com/holochain/wind-tunnel/pull/277/commits/f37d72813069d02aea4f0a5330288af66f9180b3

It does seem like this scenario is a generalized demonstration of the performance of `get_agent_activity`, through the various conditions it would find in real usage. Not sure if of the under-the-hood implementation details are really important, but I guess its not a bad thing to give any reviewers some hints to explain what they're seeing.

ThetaSinner:
Code changes look good, thank you.

I've just given running this a go. The dashboard looks good but I'm not able to get the summarizer to run. So I'm seeing data in Influx

image


But for the same run_id locally I'm seeing 

```
[2025-09-26T09:38:52Z DEBUG holochain_summariser] Loading summary from run_summary.jsonl
[2025-09-26T09:38:52Z DEBUG holochain_summariser] Selected summary for write_get_agent_activity (80fc1af2a7172fe34aee5ff222cfd2e8a09b7190f05722fd508c3d24199ddee7): RunSummary { run_id: "vXoQ-HKmaqOM68Bf4VEEG", scenario_name: "write_get_agent_activity", started_at: 1758879228, run_duration: Some(60), peer_count: 2, peer_end_count: 2, assigned_behaviours: {"get_agent_activity": 1, "write": 1}, env: {}, wind_tunnel_version: "0.5.0-dev.0" }
[2025-09-26T09:38:52Z DEBUG holochain_summariser::query] Querying: ReadQuery { queries: ["SELECT value, agent FROM \"windtunnel\".\"autogen\".\"wt.custom.write_get_agent_activity\" WHERE run_id = 'vXoQ-HKmaqOM68Bf4VEEG'"] }
[2025-09-26T09:38:52Z DEBUG reqwest::connect] starting new connection: http://localhost:8087/
Error: Agent activity summary

Caused by:
    0: Load write_get_agent_activity data
    1: No series in result: Object {"statement_id": Number(0)}
```

It might just be that I've missed a step! I've run with host metrics enabled and uploaded everything at the end of the run with the local upload script. I'd expect that seeing data in the dashboard would be enough to run the summarizer.

If you've run it successfully, them I'm happy, I just want to check it's working okay.

mattyg:
@ThetaSinner Nope you're right the summary was broken. Fixed in https://github.com/holochain/wind-tunnel/pull/277/commits/5a1ef400038e8b38aa7e6cdca606123cd6b0f790

PR #282, opened 2025-09-24
Feat/write validated must get agent activity

mattyg:
### Summary
Scenario to measure time to run a zome call that uses `must_get_agent_activity` in validation, as the target agent's chain length increases.

### TODO:


mattyg:
@coderabbit review

PR #285, opened 2025-09-29
chore: 222 refactor nix flake

### LINKED ISSUE #222: SPIKE: Refactor Nix flake
The Nix code is currently quite clean and easy to understand, so this isn't a big thing, but there are some minor improvements that I think it would be good to look at:

1. Currently, the Rust toolchain is defined in https://github.com/holochain/wind-tunnel/blob/main/nix/modules/rust.nix. I want to move the definition to a standard `rust-toolchain.toml` file and use the `fromRustupToolchainFile` function from the Nix Rust overlay to read the channel, version, components, etc.
2. Currently, static checks can be run locally and in the CI with the script file: https://github.com/holochain/wind-tunnel/blob/main/scripts/checks.sh. I propose moving to use https://github.com/cachix/git-hooks.nix, which has the existing checks and more, and allows us to run the checks automatically on git commit and manually with `nix flake check`.
3. Look at moving all the scripts to Nix packages, allowing us to run them with `nix run .#scipt-name`. If we want to allow people to still run them manually without Nix then we could keep the scripts as separate files but still package them in Nix by reading from these files.

veeso:
### Summary

- **chore: refactor nix flake**
- **fix: rust-toolchain path**
- **fix: refactored nix scripts and ci**

closes #222

### TODO:

## Summary by CodeRabbit

- New Features
  - Unified developer commands: checks, formatting, tests, and job generation now run via the package manager runtime.
  - Tooling: added consistent Rust toolchain config and new TOML/YAML formatting config.

- Refactor
  - CI/workflows consolidated to the unified runtime command flow; test/smoke-job invocation simplified.
  - Formatting helpers added and formatting orchestration consolidated; some legacy formatter scripts removed.

- Chores
  - .gitignore updated and environment script header standardized.
  - Minor robustness and whitespace fixes across configs.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -197,7 +197,7 @@
           check-nix-fmt = pkgs.writeShellApplication {
             name = "check-nix-fmt";
             runtimeInputs = [
-              pkgs.nixpkgs-fmt
+              pkgs.nix
             ];
             text = ''
               set -euo pipefail
@@ -225,7 +225,7 @@
           check-rust-fmt = pkgs.writeShellApplication {
             name = "check-rust-fmt";
             runtimeInputs = [
-              pkgs.rustPlatform.bindgenHook
+              config.rustHelper.rust
             ];
             text = ''
               set -euo pipefail
@@ -239,10 +239,16 @@
           check-rust-lint = pkgs.writeShellApplication {
             name = "check-rust-lint";
             runtimeInputs = [
+              config.rustHelper.rust
+              pkgs.perl
+              pkgs.gnumake
+              pkgs.cmake
               pkgs.rustPlatform.bindgenHook
             ];
             text = ''
               set -euo pipefail
+
+              export LIBCLANG_PATH="${pkgs.libclang.lib}/lib"
 
               # shellcheck disable=SC1091
               source ${./scripts/checks.sh}
@@ -297,12 +303,19 @@
             runtimeInputs = [
               pkgs.go
               pkgs.rustPlatform.bindgenHook
+              config.rustHelper.rust
+              pkgs.perl
+              pkgs.gnumake
+              pkgs.cmake
               pkgs.shellcheck
               pkgs.taplo
               pkgs.yamlfmt
+              pkgs.statix
             ];
             text = ''
               set -euo pipefail
+
+              export LIBCLANG_PATH="${pkgs.libclang.lib}/lib"
 
               # shellcheck disable=SC1091
               source ${./scripts/checks.sh}
@@ -313,7 +326,7 @@
           format-rust = pkgs.writeShellApplication {
             name = "format-rust";
             runtimeInputs = [
-              pkgs.rustPlatform.bindgenHook
+              config.rustHelper.rust
             ];
             text = ''
               set -euo pipefail
@@ -355,10 +368,9 @@
           format-all = pkgs.writeShellApplication {
             name = "format-all";
             runtimeInputs = [
-              pkgs.nixpkgs-fmt
-              pkgs.rustPlatform.bindgenHook
               pkgs.yamlfmt
               pkgs.taplo
+              config.rustHelper.rust
             ];
             text = ''
               set -euo pipefail
@@ -384,10 +396,16 @@
           rust-unit-tests = pkgs.writeShellApplication {
             name = "rust-unit-tests";
             runtimeInputs = [
+              config.rustHelper.rust
+              pkgs.perl
+              pkgs.gnumake
+              pkgs.cmake
               pkgs.rustPlatform.bindgenHook
             ];
             text = ''
               set -euo pipefail
+
+              export LIBCLANG_PATH="${pkgs.libclang.lib}/lib"
 
               cargo test --workspace --all-targets
             '';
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -147,13 +179,252 @@
...
+            ];
+            text = ''
+              set -euo pipefail
+
+              # shellcheck disable=SC1091
+              source ./scripts/tests.sh
+              
+              smoke_test "$@"
+            '';
+          };
```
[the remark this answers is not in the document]
veeso: cd ${toString ./.}

does this work to cd to the project root?

lp-tool/lp-tool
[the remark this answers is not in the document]
veeso: No, it shouldn't 😅

flake.nix
```diff
@@ -4,6 +4,8 @@
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-25.05";
     flake-parts.url = "github:hercules-ci/flake-parts";
+    git-hooks.url = "github:cachix/git-hooks.nix";
+    systems.url = "github:nix-systems/default";
```
ThetaSinner, line 8: What is this needed for?

flake.nix
```diff
@@ -23,10 +25,15 @@
...
-  outputs = inputs@{ flake-parts, crane, rust-overlay, nixpkgs, ... }: flake-parts.lib.mkFlake { inherit inputs; } ({ flake-parts-lib, ... }: {
+  outputs = inputs@{ flake-parts, git-hooks, crane, rust-overlay, nixpkgs, ... }: flake-parts.lib.mkFlake { inherit inputs; } {
+    imports = [
+      git-hooks.flakeModule
+    ];
+
     systems = builtins.attrNames inputs.holonix.devShells;
     perSystem = { inputs', pkgs, system, config, ... }:
       let
+        flake-parts-lib = inputs.flake-parts.lib;
```
ThetaSinner, line 36: Extracted variable does not shorten the statement, is this needed?

rust-toolchain.toml
ThetaSinner, line 1: It seems that Nix is not loading this file? See https://github.com/holochain/wind-tunnel/blob/main/nix/modules/rust.nix#L11-L16

scripts/bump.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-current_version=$(tomlq -r '.workspace.dependencies.wind_tunnel_core.version' Cargo.toml)
+current_version=$(tq -r '.workspace.dependencies.wind_tunnel_core.version' Cargo.toml)
```
ThetaSinner, line 3: Same binary provided by https://search.nixos.org/packages?channel=25.05&show=tomlq&query=tomlq but I don't find this clearer, is `tomlq` such a long name? Yes if you're typing it but in a script perhaps it was fine?

flake.nix
```diff
@@ -54,19 +66,34 @@
...
         ];
 
+        pre-commit = {
+          check.enable = true;
+          settings = {
+            hooks = {
+              nixpkgs-fmt.enable = true;
+              statix.enable = true;
+              shellcheck.enable = true;
+              rustfmt.enable = true;
```
ThetaSinner, line 76: There are two versions of Rust imported into the shell now, 1.87 and 1.88. I assume that's because of this. Is there any way to provide a toolchain to the pre-commit checks?

ThetaSinner (commented):
Took me a bit of time to get my head around what's changing here. I think the changes looks good, though I think a review from @cdunster would be valuable here. I have left some possible polish comments, and one about the toolchain file that I think is important.

flake.nix
cdunster: AFAIK, @ThetaSinner is correct and `git-hooks` will use `rustfmt` from `nixpkgs` but it can be overridden by setting:

```nix
rustfmt.packageOverrides.cargo = rustToolchain;
rustfmt.packageOverrides.clippy = rustToolchain;
```

But then `rustToolchain` needs to be defined as:

```nix
let
  rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
in
```

flake.nix
```diff
@@ -54,19 +64,34 @@
...
+
         devShells =
           let
             # The packages required in most devShells
             commonPackages = [
               pkgs.cmake
               pkgs.gomplate
+              pkgs.netcat-gnu
               pkgs.perl
+              pkgs.pre-commit
```
cdunster, line 89: Do we need `pre-commit`? `git-hooks.nix` should take care of `pre-commit` for us without us having to directly interact with it.

flake.nix
```diff
@@ -147,13 +173,254 @@
...
           };
+          check-scripts = pkgs.writeShellApplication {
+            name = "check-scripts";
+            runtimeInputs = [
+              pkgs.shellcheck
+            ];
+            text = ''
+              set -euo pipefail
+
+              # shellcheck disable=SC1091
```
cdunster, line 184: Based on https://www.shellcheck.net/wiki/SC1091, could this not be:

```suggestion
              # shellcheck source=scripts/checks.sh
```

Same with the others like this below.

cdunster (changes requested):
Some comments and questions, but the main issue for me is the `kitsune_host` comment in tests.sh.

flake.nix
```diff
@@ -147,13 +173,254 @@
...
+              source ./scripts/checks.sh
```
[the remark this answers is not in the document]
veeso: It doesn't work, and I've never found any way to get it working without that shell check...

PR #286, opened 2025-10-01
ci: Upload run_summary artifact to github ci

### LINKED ISSUE #270: Publish the unified scenario summaries
Once the work on #269 has been done, we will have a single summary for each scenario run. These summaries should be published/stored somewhere so that we can compare the latest run with the previous ones.

veeso:
### Summary

- Upload the run_summary.jsonl to Github ci artifacts.

closes #270

### TODO:

## Summary by CodeRabbit

* **Chores**
  * CI now uploads the summariser run report as an artifact after the summarisation step, improving visibility and retrieval.
  * Upload configured to skip gracefully when no report is produced.
  * No changes to product behavior or public interfaces; impacts CI reporting and traceability only.

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -182,3 +182,10 @@ jobs:
           RUN_SUMMARY_PATH: ./run_summary.jsonl
           RUST_LOG: debug
         run: nix develop --command cargo run --release --bin holochain-summariser
+
+      - name: Upload run summary
+        uses: actions/upload-artifact@v4
+        with:
+          name: summariser-report
+          if-no-files-found: ignore
+          path: ./summariser-report-*.json


.github/workflows/nomad.yaml
```diff
@@ -178,3 +178,9 @@ jobs:
           RUN_SUMMARY_PATH: ./run_summary.jsonl
           RUST_LOG: debug
         run: nix develop --command cargo run --release --bin holochain-summariser
+
+      - name: Upload run summary
+        uses: actions/upload-artifact@v4
+        with:
+          name: run_summary
+          path: ./run_summary.jsonl
```
cdunster, line 186: Is this the correct artefact? It should be the reports that are uploaded, not the summary.

cdunster (changes requested):

ThetaSinner (approved):

veeso:
@ThetaSinner I think it was discussed at the refinement, but atm we don't have a platform I guess where to store these artifacts.

@cdunster Unfortunately no. I have to create a task to fix the run, but atm it's impossible to run the summariser on the nomad ci, because some metrics are always missing.

ThetaSinner:
We're the only ones that are going to change that. We can commit to Git, use GitHub releases, request a storage provider from TechOps. Again, I've approved and I think this is a good step to take - but I think the task needs a follow up. The eventual plan I had for this data was to generate an HTML report which we then would want to put in source control.

PR #295, opened 2025-10-15
ci: fix summariser job in Nomad workflow

### LINKED ISSUE #289: Nomad workflow fails to summarize the run due to missing metrics
Currently, as it's possible to see in this run

https://github.com/holochain/wind-tunnel/actions/runs/18275472136/job/52027509510

the summariser always fails to run due to some missing metrics.

It's not clear whether the scenario is not running correctly or if those metrics are no longer generated.

cdunster:
### Summary

closes #289

It seems that some of the scenarios are not working as expected and so are not reporting some metrics, either due to an error with the metric-reporting logic or the scenarios themselves not working and so the metrics that report the number of successes correctly report 0 which breaks the report generation. Due to the issues being part of the scenarios themselves and each one having a different problem, I have just disabled the offending scenarios for now and created #300 to track the fixing of the scenarios. I did this so that this PR can be merged first as it has some other necessary fixes.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Reports are now saved with precise UTC timestamped filenames for clearer identification.

* **Chores**
  * CI workflow defaults updated to a newer unstable runtime and some test scenarios temporarily disabled.
  * Added a longer job timeout to CI to reduce premature cancellations.
  * Development environment expanded to include cryptography and TOML tooling.
  * Configuration token parsing in scripts streamlined for more reliable retrieval.

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -27,29 +27,29 @@ jobs:
           - local_signals
           - remote_call_rate
           - remote_signals
-          # - single_write_many_read TODO: Fix this scenario and re-enable it
-          # - two_party_countersigning TODO: Fix this scenario and re-enable it
+          # - single_write_many_read TODO: Fix this scenario and re-enable it (see https://github.com/holochain/wind-tunnel/issues/301)
+          # - two_party_countersigning TODO: Fix this scenario and re-enable it (see https://github.com/holochain/wind-tunnel/issues/302)
           - write_query
-          # - write_read TODO: Fix this scenario and re-enable it
+          # - write_read TODO: Fix this scenario and re-enable it (see https://github.com/holochain/wind-tunnel/issues/303)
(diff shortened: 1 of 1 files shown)


summariser/src/main.rs
```diff
@@ -79,7 +79,10 @@ async fn main() -> anyhow::Result<()> {
     if summary_outputs.is_empty() {
         log::warn!("No reports were generated");
     } else {
-        let report = File::create_new(format!("summariser-report-{:?}.json", Utc::now()))?;
```
ramazankoc1989-cyber: use std::fs::File;
use chrono::Utc;

let timestamp = Utc::now().format("%Y-%m-%d_%H-%M-%S"); // Dosya adı için uygun format
let filename = format!("summariser-report-{}.json", timestamp);
let report = File::create_new(&filename)?;

PR #312, opened 2025-10-22
feat: split wait for jobs per scenario

### LINKED ISSUE #296: Split the wait_for_jobs job into a job per scenario
Currently, there is a job in the Nomad CI workflow that waits for the Nomad allocations to complete and reports a success if they all complete. See:

https://github.com/holochain/wind-tunnel/blob/8329ff39cc8b9e5d4c735cfe11bcbdede0e99fde/.github/workflows/nomad.yaml#L144-L191

This setup means that the whole job fails if a single allocation fails or if generating a summary for a single scenario fails. We then have to dive into the logs to find which scenario failed, and the job fails fast, so we don't know if the following scenarios were successful or not.

Instead, we should use a matrix to report the status of each allocation individually. This matrix can also run jobs in parallel.

My suggestion would be to use JSON instead of CSV files for the allocation ID and run summary stuff, and then use GitHub's `fromJSON()` function to generate a matrix from the JSON.

mattyg:
### Summary
Resolves #296

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Standardized CI job naming and run identifiers to use environment-based JOB_NAME and RUN_ID for consistent tracking.
  * Shifted from global allocation aggregation to per-job allocation handling; artifacts, alloc_ids, summaries, and upload names now use per-job alloc_ids and JOB_NAME.
  * Added steps to persist, collect, and rehydrate matrix outputs (collect-github-matrix-outputs) and a per-job wait-for-job step.
  * Removed the legacy per-allocation wait dispatcher and its wait_for_jobs entry.

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -230,5 +230,4 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: summariser-report
-          if-no-files-found: ignore
           path: ./summariser-report-*.json


.github/workflows/nomad.yaml
```diff
@@ -192,5 +171,4 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: summariser-report
-          if-no-files-found: ignore
```
mattyg: I assume this was only added to allow the summarizer to proceed.

Now a job is considered failed if no summary is generated.

.github/workflows/nomad.yaml
```diff
@@ -131,41 +130,81 @@ jobs:
...
+
+          echo "result=$(cat out.json)" >> $GITHUB_OUTPUT
+
+    outputs:
+      result: "${{ steps.restructure.outputs.result }}"
+
+  wait-for-job:
+    name: Wait for job
+    runs-on: [self-hosted, wind-tunnel]
+    needs: collect-github-matrix-outputs
```
cdunster: Is it possible to "need" another matrix job? It would be really nice if for example the `Wait for job (app_install_minimal)` job can need the `Run app_install_minimal` job directly so then if only that job fails then we can rerun just that scenario. I'm not sure if this is possible with GitHub matrix jobs though.

.github/workflows/nomad.yaml
```diff
@@ -131,41 +130,81 @@ jobs:
...
+        uses: cloudposse/github-action-matrix-outputs-write@v1
+        with:
+          matrix-step-name: ${{ github.job }}
+          matrix-key: ${{ env.RUN_ID }}
+          outputs: |-
+            alloc_ids: ${{ steps.run-nomad-job.outputs.alloc_ids }}
+            run_id: ${{ env.RUN_ID }}
+            job_name: ${{ steps.run-nomad-job.outputs.job_name }}
+
+  collect-github-matrix-outputs:
```
jost-s: Why is this a separate job and not the last step in `run-scenarios`?

jost-s (approved):

.github/workflows/nomad.yaml
mattyg: This aggregates the results of the GITHUB_OUTPUT from each matrix job into a single data structure, so we can use it to define the next matrix.

.github/workflows/nomad.yaml
```diff
@@ -131,41 +130,81 @@ jobs:
...
+      - name: Restructure merged output
+        id: restructure
+        run: |
+          echo '${{ steps.read.outputs.result }}' | jq -c -R 'fromjson |
+            (.alloc_ids | keys) as $keys |
+            [ $keys[] as $key |
+              {
+                job_name: .job_name[$key],
+                run_id: .run_id[$key],
+                alloc_ids: .alloc_ids[$key]
```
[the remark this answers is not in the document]
cdunster: I think that this is a `jq` query so `.job_name[$key]` is actually indexing the array called `job_name` with the index of `$key`, I don't think it's a name :slightly_smiling_face:

PR #325, opened 2025-11-04
fix: Nomad CI issues

cdunster:
### Summary

This is a collection of fixes to the Nomad CI pipeline to reduce the flakiness of it. Each of the points listed below could probably be a separate PR but that seems overkill because they're all so small so I've grouped them in one PR. The things what were fixed are:

- The `two_party_countersigning` scenario was only running with 2 agents. AFAIK, this wouldn't work because then there is no witness and so the scenario will always fail. Adding a new agent with the `participate` behaviour means there is now a witness and it works.
- We were only waiting for a single Nomad client to run the `two_party_countersigning` scenario even though we required 2 and now 3 clients. This has been increased.
- All of the Summariser reports had the same artefact name which caused conflicts and failed the uploading of the report as an artefact.
- Timeouts when starting a conductor. Normally, the conductor starts in about 3 seconds. Sometimes it takes more like 10 seconds and so it was timing out and there was no conductor running when trying to install the scenario hApp
- Waited endlessly for free Nomad clients. At some point we only had 2 clients online and we needed 3 to run a scenario, the CI ran forever so I've now added a timeout.
- Make sure that the JSON used by the CI to persist the Nomad allocation IDs, is a string because [there was one weird run](https://github.com/holochain/wind-tunnel/actions/runs/19029796045/job/54341848742) where the allocation ID was a valid number and so it was used as a number when requesting the status from Nomad which failed because Nomad didn't know of an allocation ID `6.9944E+33`

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased conductor startup timeout to improve reliability during initialization.

* **Chores**
  * Enhanced test infrastructure with expanded scenario coverage and improved result reporting.

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -148,9 +148,9 @@ jobs:
           matrix-step-name: ${{ github.job }}
           matrix-key: ${{ env.RUN_ID }}
           outputs: |-
-            alloc_ids: ${{ steps.run-nomad-job.outputs.alloc_ids }}
-            run_id: ${{ env.RUN_ID }}
-            job_name: ${{ steps.run-nomad-job.outputs.job_name }}
+            alloc_ids: "${{ steps.run-nomad-job.outputs.alloc_ids }}"
+            run_id: "${{ env.RUN_ID }}"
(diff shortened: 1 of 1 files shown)


veeso (approved):
Nice

PR #327, opened 2025-11-05
Feat: Wind Tunnel summary visualiser

### LINKED ISSUE #311: Visualise the Summariser reports
After #270 we have the summariser reports uploaded as artefacts after every run of the [Nomad action](https://github.com/holochain/wind-tunnel/actions/workflows/nomad.yaml). These reports are JSON files that contain information about a particular scenario run. Although the information is there, pure JSON is not easy to understand and compare, therefore, we want to be able to visualise the reports as static web pages.

This issue is about generating a single web page from a single report and so hosting the pages and the data is out of scope.

AC:
- [x] Provided with a summariser report as a JSON file of an array of many scenario runs, a new HTML document can be generated from this list of runs.
- [x] The generated HTML file is self-contained and thus can be opened locally in a browser.
- [x] Each scenario will have a custom template with a description and a list of widgets to include. The tool does not need to be super generic over any data.
- [x] This issue is a first pass and so can exclude any irrelevant or less-relevant data such as multiple CPU cores and network interfaces.
- [x] The data from the report is human readable and user friendly, this is subjective but at the very least:
    - [x] There must be a short description of what the scenario does.
    - [x] The timestamp of the run must be prominently displayed, either in local time or with the timezone clearly labelled.
    - [x] There must be graphs with axis labels and units showing the data over the runtime of the scenario run.
    - [x] The version of Wind Tunnel must be displayed.
- [x] Add some tests for the UI, at the very least it should test that the tool runs but it would be nice to have tests for layout and data.

Out of scope:
- Running this new HTML page generator tool - for now this will be done locally as and when needed.
- Dynamically retrieving the data from a URL etc. - for now the data will be provided when generating the HTML page with this tool.
- Displaying multiple reports/runs - for now only one report must be displayed
- Displaying the differences between runs or reports - for now only a single report must be displayed as the diffing will happen in #307.
- Hosting the page - for now this tool will be used locally so the results must not be hosted anywhere.

pdaoust:
### Summary

Closes #311 . Includes the CLI plus a sample template for the `dht_sync_lag` report.

To test the tool out: **UPDATED INSTRUCTIONS 2025-11-11**

1. In the project root, go `nix develop`
3. run `scenario-visualiser/generate.sh ../summariser/test_data/3_summary_outputs/dht_sync_lag-3a1e33ccf661bd873966c539d4d227e703e1496fb54bb999f7be30a3dd493e51.json > out.html`
4. open `out.html` in your browser

### TODO:

- [x] tests to ensure the tool works

## Summary by CodeRabbit

* **New Features**
  * Wind Tunnel Summary Visualiser: generates HTML report pages with interactive trend graphs, mean/std/max widgets, delta displays, and a DHT Sync Lag scenario view with comprehensive metric panels.

* **Documentation**
  * Added visualiser README, scenario template guidance, and TODO/planning notes.

* **Tests**
  * Added smoke-test script and test data to verify generated HTML output for scenarios.

* **Chores**
  * CI smoke-test entry added and .gitignore updated to ignore generated HTML.

### DIFF
--- a/summary-visualiser/assets/windTunnel.js
+++ b/summary-visualiser/assets/windTunnel.js
@@ -1,0 +1,100 @@
+window.createTrendGraph = function (svgId, trendData, meanValue, windowDuration, yUnit) {
+    // Note: if you change the left margin value,
+    // make sure that the CSS for `.no-trend-graph` in `page.html.tmpl`
+    // gets updated accordingly.
+    const margin = { top: 10, right: 20, bottom: 20, left: 60 };
+    const pointWidth = 10; // Width per data point
+    const width = (trendData.length * pointWidth);
+    const height = 120;
+
+    const svg = d3.select(`#${svgId}`)
+        .attr("width", width + margin.left + margin.right)
+        .attr("height", height + margin.top + margin.bottom)
+        // The guts of the graph go into this `<g>` wrapper,
+        // in order to make the coordinates of the trend data and axis label easy to work with --
+        // they're all zero-referenced to the top-left corner of the graph.
+        // It's then moved to the right position in the SVG to make room for the legends.
+        .append("g")
+        .attr("transform", `translate(${margin.left},${margin.top})`);
+
+    const maxVal = d3.max(trendData);
+
+    // Map time window values to the x range of the graph.
+    const x = d3.scaleLinear()
+        .domain([0, trendData.length - 1])
+        .range([0, width]);
+
+    // Map data point values to the y range of the graph.
+    const y = d3.scaleLinear()
+        .domain([0, maxVal])
+        .range([height, 0]);
+
+    // The trend line.
+    const line = d3.line()
+        .x((d, i) => x(i))
+        .y(d => y(d));
+    svg.append("path")
+        .datum(trendData)
+        .attr("class", "trend-line")
+        .attr("d", line);
+
+    // The area under the trend line.
+    const area = d3.area()
+        .x((d, i) => x(i))
+        .y0(height)
+        .y1(d => y(d));
+    svg.append("path")
+        .datum(trendData)
+        .attr("class", "trend-area")
+        .attr("d", area);
+
+    // Draw the mean line if provided.
+    if (meanValue !== null) {
+        svg.append("line")
+            .attr("class", "mean-line")
+            .attr("x1", 0)
+            .attr("x2", width)
+            .attr("y1", y(meanValue))
+            .attr("y2", y(meanValue));
+    }
+
+    // Y-axis labels -- always zero on the bottom and the max data point value on the top.
+    svg.append("text")
+        .attr("class", "axis-label")
+        .attr("x", -5)
+        .attr("y", 0)
+        .attr("text-anchor", "end")
+        .attr("alignment-baseline", "middle")
+        .text(`${maxVal}${yUnit || ""}`);
+
+    svg.append("text")
+        .attr("class", "axis-label")
+        .attr("x", -5)
+        .attr("y", height)
+        .attr("text-anchor", "end")
+        .attr("alignment-baseline", "middle")
+        .text(`0${yUnit || ""}`);
+
+    // X-axis labels -- 0 seconds at the start,
+    // and the number of data points × the time window duration at the end.
+
+    // Extract numeric and non-numeric parts of windowDuration.
+    const match = windowDuration.match(/^(\d+)(.*)$/);
+    const durationValue = parseInt(match[1]);
+    const durationUnit = match[2];
+    const totalDuration = durationValue * trendData.length;
+
+    svg.append("text")
+        .attr("class", "axis-label")
+        .attr("x", 0)
+        .attr("y", height + 15)
+        .attr("text-anchor", "middle")
+        .text(`0${durationUnit}`);
+
+    svg.append("text")
+        .attr("class", "axis-label")
+        .attr("x", width)
+        .attr("y", height + 15)
+        .attr("text-anchor", "middle")
+        .text(`${totalDuration}${durationUnit}`);
+};
--- a/summary-visualiser/generate.sh
+++ b/summary-visualiser/generate.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+script_dir=$(dirname "$0")
+
+# Get all the helper templates and make them available to gomplate
+# via their unadorned filename
+# (e.g., `delta` not `templates/helpers/delta.html.tmpl`).
+template_args=()
+for template_dir in "helpers" "scenarios" ; do
+    for helper_file in "$script_dir/templates/$template_dir/"*.html.tmpl; do
+        [ -e "$helper_file" ] || continue
+        helper_name=$(basename "$helper_file" .html.tmpl)
+        template_args+=(-t "$helper_name=$helper_file")
+    done
+done
+
+# This command does three things worth noting:
+# * Accept either a filename or stdin
+# * Add the JS file as a helper template so that it can be inlined into the page
+# * Set up a helper plugin to test whether a template exists for a given scenario
+#   (note: this requires `SCENARIO_TEMPLATES_DIR` to be set)
+SCENARIO_TEMPLATES_DIR="$script_dir/templates/scenarios" gomplate \
+    -c .="${1:-stdin:///in.json}" \
+    -t js="$script_dir/assets/windTunnel.js" \
+    "${template_args[@]}" \
+    --plugin scenario_template_exists="$script_dir/scenario_template_exists.sh" \
+    -f "$script_dir/templates/page.html.tmpl"
--- a/summary-visualiser/templates/scenarios/dht_sync_lag.html.tmpl
+++ b/summary-visualiser/templates/scenarios/dht_sync_lag.html.tmpl
@@ -1,0 +1,344 @@
+{{/* The description is a 'partial' that gets passed to the `scenarioSummary` template.
+We could pass it as a string literal instead, but this is nicer
+because your code editor's syntax highlighter can make it look more legible. */}}
+{{- define "description" }}
+<p>Measure lag time between an agent publishing data and other peers being able to see it. This scenario has two roles:</p>
+
+<ul>
+    <li><b><code>write</code></b>: A simple job that just creates entries with a timestamp field. Those entries are linked to a known base hash. For each write, the metric <code>ws.custom.dht_sync_sent_count</code> is incremented.</li>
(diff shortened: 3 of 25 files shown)


ThetaSinner:
AC not met, the scenario is missing a description

image

ThetaSinner:
Standard notation for standard deviation would be `(SD = <number>)` rather than using `STD`

image

ThetaSinner:
Beyond my input above, I'm not feeling like I'm being listened to about keeping this simple. We've had conversations across a few threads in MM but this is still an NPM package which means we have to introduce Node.js in the Nix flake, we still have to update dependencies for it, and there is actually quite a bit of code here to dig through too.

Just looking at the `cli.js`, that looks to be confused between dealing with templates and being a bundler. The CSS is being injected as a template variable but it's also referenced in the page? `<link rel="stylesheet" href="./dist/windTunnel.css" />`. Either it should be templated with a hard-coded path, or it should be referenced by its source and let Rollup deal with it. Not both.

I could dig deeper and try to make sense of everything but this is already reading like the complexity is biting and I hope that one example motivates that.

We already have `nomad/scripts/generate_jobs.sh` which deals with templating. It's using the `.tpl.hcl` file to generate a `.hcl` file and it just has a small Bash script to coordinate that work. That's what I wanted this to look like and that's what I've been trying to get across.

I'm really struggling with seeing this approach reaching a point where I see it as maintainable. If I had to work on top of this code today, I would immediately be in a place of frustration because it's just hard to follow. I'm looking at `index.js` and my IDE is showing a bunch of useful warnings, and I can't start absorbing what the code is doing before addressing that.

summary-visualiser/src/scenarioTransforms.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * Truncate a 32-byte Base64 ID to its first four bytes
```
mattyg:
```suggestion
 * Truncate a 32-byte Base64 ID to its first three bytes
```

summary-visualiser/src/assets/windTunnel.js
```diff
@@ -0,0 +1,97 @@
+import { formatNumber } from "../util.js";
+
+window.createTrendGraph = function (svgId, trendData, meanValue, windowDuration, yUnit) {
+    const margin = { top: 25, right: 20, bottom: 20, left: 60 };
+    const pointWidth = 40; // Width per data point
+    const width = (trendData.length * pointWidth) - margin.left - margin.right;
+    const height = 120 - margin.top - margin.bottom;
+
+    const svg = d3.select(`#${svgId}`)
```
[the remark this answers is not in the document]
pdaoust: this is d3 though?

summary-visualiser/src/assets/windTunnel.js
```diff
@@ -0,0 +1,97 @@
...
+    const x = d3.scaleLinear()
+        .domain([0, trendData.length - 1])
+        .range([0, width]);
+
+    const y = d3.scaleLinear()
+        .domain([0, yMax])
+        .range([height, 0]);
+
+    // Create line generator
+    const line = d3.line()
```
mattyg: I think line charts were all the same width, not adjusted based on x max.

summary-visualiser/src/templates/partials/scenarioSummary.hbs
```diff
@@ -0,0 +1,34 @@
...
+        <div class="scenario-summary-details-line">
+            <div class="scenario-summary-details-label">Duration</div>
+            <div class="scenario-summary-details-value">{{number data.run_duration}}s</div>
+        </div>
+
+        <div class="scenario-summary-details-line">
+            <div class="scenario-summary-details-label">Behaviours</div>
+            <div class="scenario-summary-details-value">
+                <ul class="scenario-summary-assigned-behaviours">
+                    {{#each data.assigned_behaviours}}<li><code>{{@key}}</code>: {{plural this " agent" " agents"}}</li>{{/each}}
```
mattyg: Styling of these looks a little funky:

Image

summary-visualiser/cli.js
```diff
@@ -0,0 +1,101 @@
...
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const program = new Command();
+program
+    .name("wind-tunnel-summary-visualiser")
+    .description("A tool to generate a pretty HTML report from Wind Tunnel scenario run summary JSON.")
+    .version("0.0.1")
+    .option("-t, --template <template>", "HTML page template to insert the generated HTML into.", "page.hbs")
```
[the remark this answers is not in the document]
pdaoust: fair enough.

summary-visualiser/src/assets/windTunnel.js
pdaoust: interesting -- I grant that they look odd when they're all different lengths, but I think that it's important here to keep 10sec time units a consistent width and use the width of the whole graph to show the length of the time series. So a longer time series (e.g., 70 sec vs 20 sec) should be a longer graph. Does that make sense, or is this not useful information?

summary-visualiser/src/templates/partials/scenarioMetric.hbs
```diff
@@ -0,0 +1,4 @@
+<div class="scenario-metric {{cssClass}}">
+    <div class="scenario-metric-name">{{{name}}}</div>
```
mattyg:
```suggestion
    <div class="scenario-metric-name">{{name}}</div>
```

summary-visualiser/cli.js
```diff
@@ -0,0 +1,101 @@
...
+    .argument("<inputFile>", "The path to the input JSON.")
+    .argument("<outputFile>", "the path to the HTML file you want to create.");
```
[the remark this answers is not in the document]
mattyg: I was thinking dist/index.html

summary-visualiser/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+// Load all the partials and make them available to the template.
+const partials = fs.readdirSync(partialsFolderPath);
+for (const p of partials) {
+    const partialName = p.replace(".hbs", "");
+    const partialContent = fs.readFileSync(path.join(partialsFolderPath, p), { encoding: "utf8" });
+    handlebars.registerPartial(partialName, partialContent);
+}
+
+// A few helpers
```
mattyg: I know handlebars doesn't really allow logic in templates by-design, and the helpers are how you work around that. Would it make more sense to use a template system that allows logic?

pdaoust: Yes! I'm switching to `gomplate` which also seems logicless but has an absolute heap of helper functions. It does mean you need to do Polish notation for comparisons and arithmetic, but it means less code for us to maintain.

summary-visualiser/assets/windTunnel.js
```diff
@@ -0,0 +1,97 @@
...
+    const x = d3.scaleLinear()
+        .domain([0, trendData.length - 1])
+        .range([0, width]);
+
+    const y = d3.scaleLinear()
+        .domain([0, yMax])
+        .range([height, 0]);
+
+    // Create line generator
+    const line = d3.line()
```
mattyg: We still have that information in the x max. Idk I would do full width for everything, but I'm not sure one is better than another, so I'll defer to you.

summary-visualiser/.gitignore
```diff
@@ -0,0 +1,2 @@
+dist
+*.html
```
ThetaSinner: The instructions from the README generate the HTML to the root of the project. Suggest moving this content to the root `.gitignore`.

summary-visualiser/assets/windTunnel.js
ThetaSinner: In the output I've generated locally I'm not seeing any units on these graphs?

image

summary-visualiser/generate.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -euo pipefail
+
+SCRIPT_DIR=$(dirname "$0")
+# Read either from a filename or stdin.
+INPUT_JSON=$(cat "${1:-/dev/stdin}")
+# The scenario name is pulled from the JSON itself.
+# We'll need this later to fetch the right template.
+SCENARIO_NAME=$(echo "$INPUT_JSON" | jq -r ".run_summary.scenario_name")
```
ThetaSinner: This might work with one element, but the content of this is an array isn't it? At least that's what I have in my latest generated data:

```json
[
  {
    "run_summary": {
      "run_id": "CKYEOs9IAH5W2n4s2TyXd",
      "scenario_name": "zero_arc_create_data",
      "started_at": 1762773592,
      "run_duration": 300,
      "peer_count": 3,
      "peer_end_count": 3,
      "assigned_behaviours": {
        "zero": 1,
        "full": 2
      },
```

ThetaSinner: I think the problem here is trying to determine the scenario name in the script. That's not what I've been suggesting.

If instead there is a single top level template that contains something like this pseudo-code:

```
{foreach scenario in . }
  {if scenario.run_summary.scenario_name == "dht_sync_lag" }
    {renderDhtSyncLag scenario }
  {else}
    <p>No template for scenario {scenario.run_summary.scenario_name}</p>
  {/end}
{/end}
```

Then less work is needed in the script and the problem of handling multiple scenarios is set up ready for working with multiple and the script doesn't need adjusting later.

summary-visualiser/templates/scenarios/dht_sync_lag.html.tmpl
```diff
@@ -0,0 +1,315 @@
...
+        "description" (tmpl.Exec "description")
+        "data" .scenario.run_summary
+    ) }}
+
+    <div class="scenario-metrics">
+        {{ with .scenario.scenario_metrics }}
+            {{ with (index .create_rate.rates 0).summary_rate }}
+                {{ template "scenarioMetric" (dict
+                    "name" "Create rate"
+                    "description" "TODO"
```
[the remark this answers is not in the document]
ThetaSinner: That is as agreed, somebody who knows the details of these metrics can come and fill these in.

ThetaSinner (commented):
I definitely think this is simpler. The templates are always a bit fiddly and finding and fixing problems in them can be a pain but that was the case with handlebars too. Having a separate JS file that gets included is good, that means you get IDE support and dropping it in as a script tag means no extra tooling is needed. 

For the top level of how the script interacts with templates, I've commented on how I think that should work. I don't think the script needs to decide what's in the input JSON. Just assume it's an array and pass it to a root template.

summary-visualiser/generate.sh
pdaoust: https://github.com/holochain/wind-tunnel/pull/327/commits/2dd03f1b2e7f5a5b361eb2dabb0e24061866ce70

Note that this commit doesn't iterate through all the known scenario template names; instead, by convention, it loads up all the files in `templates/scenarios/` as helper templates, and hard-fails if one doesn't exist.

I hope this blend of convention and 'strong typing' isn't weird -- I figure the former reduces long-term maintenance and diff noise, while the latter is already true about the individual scenarios (it assumes a certain data structure and fails if it encounters something unexpected).

ThetaSinner (approved):

PR #328, opened 2025-11-06
ci: Run nomad ci workflow if there are relevant files changed

### LINKED ISSUE #292: Run Nomad CI workflow when relevant files changed
Currently, the [Nomad workflow](https://github.com/holochain/wind-tunnel/actions/workflows/nomad.yaml) doesn't run when a PR is opened because it would slow down the CI too much and delay getting things merged. This is mostly okay because the workflow is run once a week, so the time to find the issue should be short. However, if someone changes the workflow itself or Nomad-related files, then the workflow should be run to check that it works.

AC:
- [ ] Update the standard Test workflow that runs on PRs to generate the Nomad job spec files and perform the `nomad job verify` command on them.
- [ ] Update the PR template to check "Did I run the scenario on Nomad"

veeso:
## Summary

Automatically runs the nomad ci workflow if either the nomad workflow file or any scenario file changed

closes #292

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Added a CI checklist item requiring a successful Nomad workflow run.
  * Enabled the Nomad workflow to trigger on changes via push and pull requests.

* **New Features**
  * Test CI now runs a Nomad job validation step before smoke tests.
  * Added a reusable validation target for Nomad jobs.
  * Job generation can optionally validate produced jobs and verifies the Nomad CLI is available.

### DIFF
--- a/nomad/scripts/generate_jobs.sh
+++ b/nomad/scripts/generate_jobs.sh
@@ -4,20 +4,20 @@ set -euo pipefail
 set -x
 
 generate_job() {
-  local var_file="$1"
-  local job_file="$2"
-  gomplate \
-    -f "$TEMPLATE" \
-    -o "$job_file" \
-    -d vars="$var_file"
+    local var_file="$1"
+    local job_file="$2"
+    gomplate \
+        -f "$TEMPLATE" \
+        -o "$job_file" \
+        -d vars="$var_file"
 
-  echo "Generated job file: $job_file"
+    echo "Generated job file: $job_file"
 }
 
 validate_job() {
-  local job_file="$1"
-  nomad job validate "$job_file"
-  echo "Validated job file: $job_file"
+    local job_file="$1"
+    nomad job validate "$job_file"
+    echo "Validated job file: $job_file"
 }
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nomad.yaml
```diff
@@ -7,6 +7,18 @@ on:
...
+  push:
+    branches:
+      - main
+    paths:
+      - "nomad/**"
+      - ".github/workflows/nomad.yaml"
+  # Trigger on changes to Nomad-related files in PRs
+  pull_request:
+    paths:
+      - "nomad/**"
```
cdunster: From what I remember from the refinement meeting: we decided not to do this and only run this workflow when the workflow file itself changes as below. The other workflow (`test.yaml`) should be updated as specified in the AC to generate the Nomad job files and run `nomad job validate` for each of them on every PR.

cdunster (changes requested):
The main thing that I'm not seeing is the point from the AC:

> Update the standard Test workflow that runs on PRs to generate the Nomad job spec files and perform the `nomad job verify` command on them.

Issue #330, opened 2025-11-07
Use Rust cache for Summariser in the CI


cdunster:
The CI runs the Summariser via `cargo run`. This causes a build of the Summariser binary in every job and this job is run once for each scenario that is run.

To avoid unnecessarily rebuilding the Summariser every time we should use the `rust-cache` GitHub action that we use in other repos.

PR #331, opened 2025-11-07
ci: run the Summariser report generation once

cdunster:
### Summary

Currently, we run the Summariser once for each scenario after it finishes. This means that not only do we rebuild the Summariser once for each scenario (see #330) but we also end up with multiple Summariser reports (one for each scenario). This does mean that if the Summariser fails on a single scenario then only that scenario fails to get a report instead of the whole job failing, but I don't think that this is the intended use of the Summariser.

This PR makes it so that the Summariser is run once at the end of the run and generates a single report for all scenarios that were run.

This is a suggestion so I'm happy to reject this PR and keep the existing behaviour but then we should probably have a job that combines the reports into a single one so it can be used by the visualiser tool being developed as part of #311.

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Improved CI workflow for consolidated run-summary generation and reporting: the wait job was renamed for clarity, a dedicated run-summary job now merges allocation IDs into a single source, summarisation always runs, and the summariser report is uploaded under a unified name for easier retrieval.
  * No user-facing changes.

### DIFF
--- a/.github/workflows/nomad.yaml
+++ b/.github/workflows/nomad.yaml
@@ -228,6 +228,7 @@ jobs:
         run: nix develop --command ./nomad/scripts/ci_allocs.sh generate_run_summary alloc_ids.csv run_summary.jsonl
 
       - name: Run summariser
+        if: always()
         env:
           INFLUX_HOST: "https://ifdb.holochain.org"
           INFLUX_BUCKET: "windtunnel"


ThetaSinner (approved):
I'm not entirely sure what the best option is here really. The summarizer is coded to run all reports in parallel and it won't produce a report if any of them fail. Though it will run them all to completion before stopping with an error.

That behavior could be changed so that failed reports are just logged and the successful ones are produced.

Whether this approach performs acceptably is hard to say. Running reports in parallel is nice for shortening the total runtime when you've run scenarios for a short time - but ideally we'd run some of our scenarios over hours or days. Then it's a huge amount of metrics data being fetched into memory.

I think I agree with this change, but just trying to share a sense of what the consequences might be. I prefer what you've done here, and if we have to adjust how the summarizer handles errors and give it a `--serial` argument then those are simple and reasonable changes to make :)

PR #332, opened 2025-11-07
fix: Use `force_stop_scenario` if conductor fails to start

### LINKED ISSUE #324: Use `force_stop_scenario()` if conductor fails to start
Currently, if the Holochain conductor fails to start as part of the `agent_setup` then an error is returned. This error is logged and that agent does not get run, however the scenario does not fail. Instead, the other agents are still run to completion or if there is only one agent then the scenario just ends successfully.

Wind Tunnel was designed this way so that when running multiple agents, if a single agent fails for some unrelated reason then the whole scenario doesn't fail. Because of this, Wind Tunnel does have a `force_stop_scenario` function, see:

https://github.com/holochain/wind-tunnel/blob/68a71c1a4bf0b9e453f37e74c6db05b230622837/framework/runner/src/context.rs#L83-L90

This `force_stop_scenario` function should be called if a critical failure occurs. The conductor failing to run should be a critical failure, therefore if an error occurs when running the conductor we should call this function.

AC:
- Call the `force_stop_scenario` function if the conductor fails to start for whatever reason.
- Run the Nomad cluster CI workflow

veeso:
### Summary

Whenever setup_agent_fn fails, stop the scenario

closes #324

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent setup failures now stop further processing immediately and trigger cleanup to prevent lingering tasks.
  * Teardown is consistently attempted and logged after setup or execution errors to reduce resource leaks.
  * Conductor start failures now perform directory cleanup and force runner shutdown, improving error reporting and recovery.

### DIFF
--- a/bindings/runner/src/common.rs
+++ b/bindings/runner/src/common.rs
@@ -1,5 +1,5 @@
 use crate::context::HolochainAgentContext;
-use crate::holochain_runner::HolochainRunner;
+use crate::holochain_runner::{HolochainConfig, HolochainRunner};
 use crate::runner_context::HolochainRunnerContext;
 use anyhow::{bail, Context};
 use holochain_client_instrumented::prelude::{
@@ -17,7 +17,7 @@ use kitsune2_api::AgentInfoSigned;
 use kitsune2_core::Ed25519Verifier;
 use rand::seq::SliceRandom;
 use rand::thread_rng;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use std::{env, fs, io};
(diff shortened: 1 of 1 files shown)


ThetaSinner:
Yes, that is how the framework is set up. It permits running multiple agents on one conductor and if we don't permit failures then running at scale is difficult. The opt-in behavior for terminating a run is intentional.

If that doesn't work out the way it was intended to when we get this running across a bigger pool of nodes then it can be adjusted but I think for now, that behavior should be maintained.

bindings/runner/src/common.rs
```diff
@@ -686,38 +686,19 @@ pub fn run_holochain_conductor<SV: UserValuesConstraint>(
...
 
-    ctx.get_mut().holochain_runner =
-        ctx.runner_context()
-            .executor()
-            .execute_in_place(async move {
-                HolochainRunner::run(&config)
-                    .await
-                    .map(Some)
-                    .map_err(|mut err| {
-                        log::trace!("Error whilst starting conductor so cleaning up directory");
```
veeso: this code was a mess and couldn't even get formatted by rust fmt. Moved to a separate function.

PR #337, opened 2025-11-12
fix: validation_receipts scenario gets stuck

### LINKED ISSUE #329: `validation_receipts` scenario gets stuck in infinite doesn't end
Due to this loop in the scenario:

https://github.com/holochain/wind-tunnel/blob/2a46efdbbfef6a82d8408b64a43f2d06303ca01c/scenarios/validation_receipts/src/main.rs#L81

If validation doesn't finish for one agent then it never exits and thus the scenario runs forever.

This could be cause by not running on enough nodes. Currently, we run 2 nodes, each with 5 conductors and 1 agent on each conductor. Start by using 5+ nodes, each with one conductor.

AC:

- [ ] Add more Nomad client nodes to run the scenario
- [ ] Reduce the number of agents/conductors running on each node
- [ ] Fix the scenario so it works if the above doesn't fix it

veeso:
### Summary

- Increased required nodes to 6
- Refactored the scenario logic.

closes #329

### TODO:

## Summary by CodeRabbit

* **Chores**
  * CI reduced to run only the validation_receipts scenario; required validation nodes increased from 2 to 6.

* **Improvements**
  * Receipt handling changed to per-action completion for finer-grained tracking.
  * Added per-action waiting with timing metrics and improved completion reporting.
  * Waiting behavior can be controlled via an environment variable for flexible test runs.

### DIFF
--- a/scenarios/validation_receipts/src/main.rs
+++ b/scenarios/validation_receipts/src/main.rs
@@ -15,34 +15,31 @@ pub struct ScenarioValues {
     /// - if any of the receipts_complete are false, we are still pending
     /// - if all the receipts_complete are true, we are complete
     ///   so the action should be removed from the map
-    pending_actions: HashMap<ActionHash, HashMap<OpType, ReceiptsComplete>>,
+    pending_action: Option<(ActionHash, HashMap<OpType, ReceiptsComplete>)>,
 }
 
 impl ScenarioValues {
-    fn mut_op_complete(&mut self, action_hash: &ActionHash, op_type: String) -> &mut bool {
-        self.pending_actions
-            .get_mut(action_hash)
-            .unwrap()
-            .entry(op_type)
-            .or_default()
+    /// Get a mutable reference to the receipts_complete for the given action and op type.
+    ///
+    /// If it returns [`None`] the action hash has completed or does not exist.
+    fn mut_op_complete(&mut self, action_hash: &ActionHash, op_type: String) -> Option<&mut bool> {
+        let inner = self
+            .pending_action
+            .get_or_insert_with(|| (action_hash.clone(), HashMap::new()));
+        Some(inner.1.entry(op_type).or_default())
     }
 
-    fn mut_any_pending(&mut self) -> bool {
-        self.pending_actions.retain(|_, m| {
-            if m.is_empty() {
-                return true;
-            }
-            let mut all_complete = true;
-            for c in m.values() {
-                if !c {
-                    all_complete = false;
-                    break;
-                }
-            }
-            !all_complete
-        });
+    /// Returns whether all the actions for the given action hash are complete.
+    fn is_action_complete(&self) -> bool {
+        let Some((_hash, ops)) = self.pending_action.as_ref() else {
+            return true;
+        };
 
-        !self.pending_actions.is_empty()
+        if ops.is_empty() {
+            false
+        } else {
+            ops.iter().all(|(_, v)| *v)
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


scenarios/validation_receipts/src/main.rs
```diff
@@ -79,6 +79,16 @@ fn agent_behaviour(
...
 
     'outer: loop {
+        // check time out (twice the scenario duration)
+        if ctx
+            .scenario_duration()
+            .map(|d| d * 2)
+            .map(|max_duration| start.elapsed() >= max_duration)
+            .unwrap_or_default()
+        {
+            panic!("Timed out waiting for validation receipts to complete");
```
mattyg: I think we discussed this during refinement and unfortunately it looks like we didn't update the issue with the outcome of that discussion. My recollection is that we said we didn't want to expose the scenario duration in context, but instead refactor the scenario to remove the loop.

One option for that refactoring could be to check at the top if `pending_actions` contains any actions, and if so, check for validation receipts. If we have all the validation receipts, remove it from `pending_actions`. If `pending_actions` is empty, create another entry.

ThetaSinner (changes requested):
I think shutdown is only implemented for `Ctrl+C` at the moment, but the framework could be slightly modified to send a shutdown signal when shutting down normally. I think that's the more general solution here that doesn't let the scenario behavior try to do anything fancy with the duration.

scenarios/validation_receipts/src/main.rs
```diff
@@ -61,82 +58,81 @@ fn agent_setup(
...
+                // and mark it complete
+                reporter.add_custom(
+                    ReportMetric::new("validation_receipts_complete_time")
+                        .with_tag("op_type", set.op_type.clone())
+                        .with_tag("agent", agent.clone())
+                        .with_field("value", start.elapsed().as_secs_f64()),
+                );
+                *cur = true;
             }
         }
```
mattyg: Do we need to remove the action from `pending_actions` once we know that all the required receipts have been received?

Also do we want to track when all receipts for all op types have been received as another metric?

Also heads-up that I believe that currently the `Vec<ValidationReceiptSet>` only contains elements when we have received at least 1 ValidationReceipt for that op type. So we may also need to check here that the list length is equal to the number of op types created.

scenarios/validation_receipts/src/main.rs
```diff
@@ -61,82 +58,81 @@ fn agent_setup(
...
 
-    ctx.get_mut()
+    // get the first pending action
+    let action_hash = ctx
+        .get()
         .scenario_values
         .pending_actions
-        .insert(action_hash, HashMap::new());
+        .keys()
+        .next()
```
mattyg: Heads up that keys() isn't ordered so if there are multiple this can pick a different key each time. I assume that isn't what we want.

scenarios/validation_receipts/src/main.rs
```diff
@@ -61,82 +58,81 @@ fn agent_setup(
...
+        .cloned()
+        .expect("Cannot be empty here");
 
+    // collect validation receipts until complete
     let start = Instant::now();
+    let wait_for_all = std::env::var_os("NO_VALIDATION_COMPLETE").is_none();
```
mattyg: With the env var enabled, I'm not clear it is collecting the most useful data. On every run through the scenario we're checking if we have received validation receipts for the oldest pending action. But we may receive validation receipts out-of-order from when the actions were authored. Should we instead check *all* pending actions to see if we have received validation receipts for any of them?

That said, it looks like the only deployments of the scenario currently have this env var disabled. Should we just remove it? Or should we add another deployment for running with the var enabled? Could be a follow up issue and PR.

PR #394, opened 2025-12-09
fix: filter links by agent in write validated must_get_agent_activity zome

matthme:
### Summary

The logic of the `create_validated_sample_entry` zome function used by the `write_validated_must_get_agent_activity` scenario was missing a filtering step to filter by agent. Without it, the scenario only works properly in case of a single write peer.

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined agent activity validation to ensure records authored by the specified agent are properly filtered before further processing.
  * Ordered the agent-specific activity by timestamp and used the most recent entry to determine chain state, improving accuracy of chain head selection.

### DIFF
--- a/zomes/validated_must_get_agent_activity/coordinator/src/lib.rs
+++ b/zomes/validated_must_get_agent_activity/coordinator/src/lib.rs
@@ -52,11 +52,15 @@ fn create_sample_entries_batch(count: u64) -> ExternResult<()> {
 #[hdk_extern]
 fn create_validated_sample_entry(agent: AgentPubKey) -> ExternResult<usize> {
     // Get last created action hash
-    let mut links: Vec<Link> = get_links(
+    let links: Vec<Link> = get_links(
         GetLinksInputBuilder::try_new(chain_head_anchor()?, LinkTypes::LatestAction)?.build(),
     )?;
-    links.sort_by_key(|l| l.timestamp);
-    let chain_head_link: Link = links
+    let mut links_from_write_agent = links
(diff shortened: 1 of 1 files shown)


zomes/validated_must_get_agent_activity/coordinator/src/lib.rs
```diff
@@ -52,11 +52,15 @@ fn create_sample_entries_batch(count: u64) -> ExternResult<()> {
...
     // Get last created action hash
-    let mut links: Vec<Link> = get_links(
+    let links: Vec<Link> = get_links(
         GetLinksInputBuilder::try_new(chain_head_anchor()?, LinkTypes::LatestAction)?.build(),
     )?;
-    links.sort_by_key(|l| l.timestamp);
-    let chain_head_link: Link = links
+    let mut links_from_write_agent = links
+        .into_iter()
+        .filter(|l| l.author == agent)
```
matthme: The links are added to the [chain_head_anchor](https://github.com/holochain/wind-tunnel/blob/4cabb7b1c36254a8ae0735b9d284c6980bc13e18/zomes/validated_must_get_agent_activity/coordinator/src/lib.rs#L42) which [does not distinguish between different agents](https://github.com/holochain/wind-tunnel/blob/4cabb7b1c36254a8ae0735b9d284c6980bc13e18/zomes/validated_must_get_agent_activity/coordinator/src/lib.rs#L11). Consequently, without filtering we may get links from different agents here than we would want to.

PR #397, opened 2025-12-10
feat: check summary html on pre commit

pdaoust:
### Summary

This PR adds a Git pre-commit hook that does two things:

* adds an HTML lint step to the smoke test (using html-tidy, the most minimal and dependency-free linter I could find)
* adds the smoke test to a Git pre-commit hook that runs on any change to any file in `summary-visualiser/`

I didn't want to narrow the hook's file filter down further (e.g., just to `.html.tmpl` files), because almost any component in that folder could result in changes to the generated HTML.

### TODO:

## Summary by CodeRabbit

* **Chores**
  * Improved developer tooling and pre-commit checks to run a smoke test for the summary visualiser.
  * Added common developer utilities for HTML linting and text processing.
  * Updated configuration signature to include an additional helper parameter for improved package composition.
  * Switched the visualiser test to a packaged runtime for more reliable execution.

* **Bug Fixes**
  * Improved error reporting for missing test resources and redirected failures to stderr.
  * Added a final HTML lint step to catch formatting and markup issues.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -455,8 +455,8 @@
               name = "summary-visualiser-smoke-test";
               runtimeInputs = [
                 pkgs.coreutils
+                pkgs.gnugrep
                 pkgs.gomplate
-                pkgs.grep
                 pkgs.html-tidy
               ];
               text = ''


flake.nix
```diff
@@ -300,6 +308,21 @@
               check_yaml_fmt
             '';
           };
+          check-scenario-visualiser-html = pkgs.writeShellApplication {
```
pdaoust: This, plus the addition of the smoke test to `scripts/checks.sh`, might be a bit of cargo-culting -- I saw that all manner of other things were added to these two places, so there may be value in adding this one too. No idea if that's true; I don't need it for my own purposes.

flake.nix
```diff
@@ -85,6 +85,13 @@
               };
               taplo.enable = true;
               yamlfmt.enable = true;
+              check-summary-visualiser-html = {
+                enable = true;
+                name = "check-summary-visualiser-html";
+                entry = "${config.packages.check-summary-visualiser-html}/bin/check-summary-visualiser-html";
```
mattyg:
```suggestion
                entry = "nix run .#summary-visualiser-smoke-test`
```

Then it just works and you don't need the `check-summary-visualiser-html`writeShellScriptBin

flake.nix
```diff
@@ -312,6 +362,8 @@
               pkgs.shellcheck
               pkgs.taplo
               pkgs.yamlfmt
+              pkgs.gomplate
+              pkgs.html-tidy
```
mattyg: Doesn't look like the check_all.sh script was changed, so these shouldn't be needed.

summary-visualiser/test.sh
```diff
@@ -45,3 +56,6 @@ smoke_test_scenario "zero_arc_create_and_read"
 smoke_test_scenario "zero_arc_create_data"
 smoke_test_scenario "zome_call_single_value"
 smoke_test_scenario "zero_arc_create_data_validated"
+
+# Finally, lint the generated HTML to ensure it's valid.
+echo "$html_output" | tidy -eq
```
mattyg:
```suggestion
echo "$html_output" | tidy -errors -quiet
```

Prefer long-form args in scripts -- more readable.

summary-visualiser/test.sh
```diff
@@ -4,6 +4,10 @@
 # shows that a template could be found for the scenario type given in the JSON.
 
 script_dir=$(dirname "$0")
+quiet=false
```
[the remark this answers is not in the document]
pdaoust: yes, this flag only suppresses messages sent to stdin -- so you always see errors.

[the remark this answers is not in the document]
pdaoust: I'll do a bit of experimentation again -- I seem to recall a lot of gobbledigook when testing the pre-commit hook before.

ThetaSinner:
How much time does this add to each commit?

Issue #335, opened 2025-11-11
Zero arc data retreival scenarios


ThetaSinner:
Following #214, there will be Wind Tunnel tests for the performance of 0-arc nodes, supported by a smaller number of full-arc nodes. What isn't checked by those tests, is how quickly 0-arc nodes can fetch data from full-arc nodes in order to see data created by remote agents.

The scenarios may use any valid method of discovering data. For example, getting links from a known base or from their agent key and then getting the target values. This should cover adding scenarios for checking:
- The zero arc nodes retrieve data created elsewhere on the network.
- The zero arc nodes retrieve data created elsewhere on the network, and the scenario app does app validation of data.

In addition, this group of scenarios should cover fetching agent chain data:
- Get agent chain data with `get_agent_activity`, where both 0-arc and full-arc nodes are creating data.
- Get agent chain data with `must_get_agent_activity` where both 0-arc and full-arc nodes are creating data.

In each of these cases, the scenario should record metrics for:
- The number of errors fetching data,
- The delay between data being authored and first being available to the 0-arc nodes.
- (activity only) The delay in seeing a new chain head, each time agent activity is successfully retrieved with a new chain head.

AC:
- Standard constraints for scenarios - must have dashboards, summariser update, summariser tests, and included in Nomad runs.

PR #398, opened 2025-12-11
feat: add mixed arc must_get_agent_activity scenario

matthme:
### Summary

Adds a `mixed_arc_must_get_agent_activity` scenario for #335 and rewrites the existing `write_validate_must_get_agent_activity` which was using the same zome and logic (modulo some zero arc specific parts).

The existing logic of `write_validate_must_get_agent_activity` has a writing behaviour that creates a batch of entries and then a link from the action of the last entry to a chain head anchor. The reading behaviour then gets the latest link from that anchor to infer the action hash of the current chain top and then makes a must_get_agent_activity call (in validation) with this action hash as the provided chain top.
The problem with this is that at the time when the link can be retrieved, not all entries of the chain may yet be retrievable. And in the next iteration of the behaviour run the chain top has already moved and the reading behaviour hardly ever actually succeeds to get the full chain up to the chain top that it specifies.

The logic is therefore adapted such that the reading behaviour keeps track of the last chain top that it successfully read and only moves on to attempting to read the next chain top once it succeeded to retrieve the full chain for for the previous chain top.

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added a mixed-agent performance scenario with run configuration and nomad vars; workspace updated to include the new scenario.
  * New and revised dashboards to visualize validation activity, chain-batch delays, open connections, zome call durations, and error metrics.
  * Enhanced summary generation for the new scenario and additional metrics included in reports.

* **Tests**
  * Added a smoke test, new snapshot, and multiple test-data fixtures for the scenario.

* **Documentation**
  * Added README documenting scenario behaviors, metrics, and run commands.


scenarios/mixed_arc_must_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+    BatchChainTop, CreateEntriesBatchInput, GetChainTopForBatchInput, SampleEntryInput,
+};
+
+const RECORD_OPEN_CONNECTIONS_PERIOD_MS: i64 = 3_000;
+const BATCH_SIZE: u32 = 10;
+
+#[derive(Debug, Default)]
+pub struct ScenarioValues {
+    write_peer: Option<AgentPubKey>,
+    batch_count: u32,
```
matthme: For each write behaviour run, a batch of 10 entries get created. This `batch_count` keeps track of how many such batches have been created so far.

scenarios/mixed_arc_must_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,288 @@
...
+    match ctx.get().scenario_values.write_peer.clone() {
+        Some(write_peer) => {
+            // If we can move on with the next batch, fetch the associated BatchChainTop
+            // ---
+            // Iteration 0: currently_pending_batch_chain_top is None, last_successfully_fetched_batch is None
+            // Iteration 1: currently_pending_batch_chain_top is Some({ batch_num: 0, ...}), last_successfully_fetched_batch is None
+            // Iteration 2.1: currently_pending_batch_chain_top is Some({ batch_num: 0, ...}), last_successfully_fetched_batch is Some(0)
+            // Iteration 2.x: currently_pending_batch_chain_top is Some({ batch_num: 1, ...}), last_successfully_fetched_batch is Some(0)
+            // Iteration 3.3: currently_pending_batch_chain_top is Some({ batch_num: 1, ...}), last_successfully_fetched_batch is Some(1)
+            // Iteration 3.x: currently_pending_batch_chain_top is Some({ batch_num: 2, ...}), last_successfully_fetched_batch is Some(1)
```
matthme: I tried to make the logic easier to follow with above inline comment - not sure whether I succeeded.

framework/summary_model/src/lib.rs
```diff
@@ -82,7 +82,7 @@ pub struct RunSummary {
     /// Note: This is only meaningful for single-conductor tests with the standard Wind Tunnel runner
     /// or with the TryCP runner. In general, each node only sees the roles it was assigned and not
     /// the roles that were assigned across the network.
-    pub assigned_behaviours: HashMap<String, usize>,
+    pub assigned_behaviours: BTreeMap<String, usize>,
```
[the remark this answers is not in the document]
matthme: That is correct. I think that is acceptable for our current level of usage.

scenarios/mixed_arc_must_get_agent_activity/src/main.rs
mattyg: I appreciate the attempt but I do not find this easier to follow 😄 . I would prefer prose.

scenarios/mixed_arc_must_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,288 @@
...
+                    .scenario_values
+                    .currently_pending_batch_chain_top = Some(batch_chain_top);
+            }
+
+            let batch_chain_top = ctx
+                .get()
+                .scenario_values
+                .currently_pending_batch_chain_top
+                .clone()
+                .expect("No currently pending batch chain top.");
```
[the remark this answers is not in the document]
mattyg: Could it be rewritten so we don't need to set the value, then get it immediately after? Would be clearer to read. I guess this is a nitpick.

scenarios/mixed_arc_must_get_agent_activity/README.md
```diff
@@ -0,0 +1,58 @@
+## mixed_arc_must_get_agent_activity
+
+### Description
+
+A scenario where zero-arc ('zero_write') and full-arc ('full_write') peers create entries in batches of 10 while other zero-arc ('zero_must_get_agent_activity') peers each repeatedly attempt to create an entry that leads to a `must_get_agent_activity` call as part of validation.
```
mattyg:
```suggestion
A scenario where zero-arc ('zero_write') and full-arc ('full_write') peers create entries in batches of 10 while other zero-arc ('zero_must_get_agent_activity') peers each repeatedly attempt to create an entry that includes a `must_get_agent_activity` call as part of its validation.
```

scenarios/mixed_arc_must_get_agent_activity/README.md
```diff
@@ -0,0 +1,58 @@
...
+You can run the scenario locally with the following command:
+
+```bash
+RUST_LOG=info cargo run --package mixed_arc_must_get_agent_activity -- --agents 6 --behaviour zero_must_get_agent_activity:3 --behaviour zero_write:1 --behaviour full_write:2 --duration 300
+```
+
+Running the scenario in a distributed manner is suggested to be done by partitioning your nodes. The first group run the command:
+
+```bash
+RUST_LOG=info cargo run --package mixed_arc_must_get_agent_activity -- --behaviour zero_must_get_agent_activity --duration 500
```
mattyg:
```suggestion
RUST_LOG=info cargo run --package mixed_arc_must_get_agent_activity -- --behaviour zero_must_get_agent_activity --duration 300
```

scenarios/mixed_arc_must_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+                SampleEntryInput {
+                    agent: write_peer.clone(),
+                    chain_top: batch_chain_top.chain_top,
+                    chain_len: batch_chain_top.chain_len,
+                },
+            );
+
+            if let Ok(()) = result {
+                // record the time difference between now and when the batch was created. Note that
+                // there is likely an accumulating lag here because writing peers write entries faster
```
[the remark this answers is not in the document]
matthme: I agree that the data we collect that way is of limited use. But I couldn't come up with a better approach that wouldn't require (centralized) synchronization between reading and writing behaviour. It seems to me like we're just confronted with an inherent constraint as to what we can reasonably measure, as a result of the fact that we need to know the chain top we're looking for ahead of time.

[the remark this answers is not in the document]
mattyg: I think its worth just noting in an issue. We may want a centralized data store for cases like this, if that's the best option we can come up with, I think.

scenarios/mixed_arc_must_get_agent_activity/src/main.rs
```diff
@@ -0,0 +1,287 @@
...
+    ctx.get_mut().scenario_values.batch_count += 1;
+
+    let agent_pub_key = ctx.get().cell_id().agent_pubkey().to_string();
+    let metric = ReportMetric::new("mixed_arc_must_get_agent_activity_entry_created_count")
+        .with_tag("agent", agent_pub_key)
+        .with_tag("behaviour", ctx.assigned_behaviour().to_string())
+        .with_field("value", ctx.get().scenario_values.batch_count * BATCH_SIZE);
+    ctx.runner_context().reporter().clone().add_custom(metric);
+
+    Ok(())
```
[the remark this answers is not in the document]
matthme: I'm not sure. Without delay we can be relatively confident that at the time when the reading behaviour wants to get the link for a certain batch for the first time it has already been created. With a lag this may not be clear anymore necessarily and it might be even more fuzzy as to what exactly we're measuring.

scenarios/mixed_arc_must_get_agent_activity/README.md
matthme: https://github.com/holochain/wind-tunnel/pull/398/commits/e1f13a061f0de06e16d23d6373602d4ed1a867e3

scenarios/mixed_arc_must_get_agent_activity/README.md
mattyg: Missing explanation of mixed_arc_must_get_agent_activity_fetch_batch_chain_top_error_count

matthme: https://github.com/holochain/wind-tunnel/pull/398/commits/da537fd160931cae990505d55afd8a5c1960f413

scenarios/write_validated_must_get_agent_activity/README.md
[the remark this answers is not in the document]
matthme: I removed the error count altogether now: https://github.com/holochain/wind-tunnel/pull/398/commits/da537fd160931cae990505d55afd8a5c1960f413

scenarios/mixed_arc_must_get_agent_activity/src/main.rs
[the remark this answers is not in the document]
matthme: Okay, introduced a delay of 5 seconds now https://github.com/holochain/wind-tunnel/pull/398/commits/da537fd160931cae990505d55afd8a5c1960f413

PR #411, opened 2025-12-15
Refactor/import index page

mattyg:
### Summary

Precurser to #412 
- Flake for local dev environment to serve gh pages site
- Extract list of summary reports to separate file that gets imported into index.md, so that we can simply prepend new elements to the list, rather than regenerating the entire index

### TODO:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ lp-tool/lp-tool
 
 # direnv temp directory
 /.direnv
+
+# Built jekyll site
+_site
--- a/README.md
+++ b/README.md
@@ -1,0 +1,10 @@
+# Wind Tunnel Github Pages Site
+
+## Development
+
+To run the site locally:
+
+1. Enter the nix shell with `nix develop`
+2. Serve the jekyll site with `jekyll serve`
+
+Note that when the site is deployed, Github's default jekyll theme 'minima' will be applied.
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,5 @@
+exclude:
+  - flake.*
+  - README.md
+  - .envrc
+  - .gitignore
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": "nixpkgs-lib"
+      },
+      "locked": {
+        "lastModified": 1765835352,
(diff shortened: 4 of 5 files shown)


index.md
```diff
@@ -1,11 +1,10 @@
...
-- [2025-12-13 00:17:59 (b3c55d5a8c91646298d9a4b5a292f66cf66130e2)](run.20251213001759.b3c55d5a8c91646298d9a4b5a292f66cf66130e2.html)
-- [2025-12-12 23:48:00 (199a5c8d028532bb6bb63237de33860c7272b508)](run.20251212234800.199a5c8d028532bb6bb63237de33860c7272b508.html)
-- [2025-12-12 22:43:56 (02876960f2e095bb8a41e0d183f72fd6396d2bae)](run.20251212224356.02876960f2e095bb8a41e0d183f72fd6396d2bae.html)
-- [2025-12-12 22:19:33 (f2763720ea0e262ebd57ee0c881c3abd4660f39c)](run.20251212221933.f2763720ea0e262ebd57ee0c881c3abd4660f39c.html)
-- [2025-12-12 22:04:48 (f2763720ea0e262ebd57ee0c881c3abd4660f39c)](run.20251212220448.f2763720ea0e262ebd57ee0c881c3abd4660f39c.html)
-- [2025-12-12 21:23:18 (d2eee3ee2c280bed525caeef246401be5f68588c)](run.20251212212318.d2eee3ee2c280bed525caeef246401be5f68588c.html)
-- [2025-12-11 14:44:45 (12dfc2be1919aafff58bea5c59b8a698c0967663)](run.20251211144445.12dfc2be1919aafff58bea5c59b8a698c0967663.html)
+
+## Canonical Runs
+{% include canonical-list.md %}
```
mattyg: Import the list partial. Now we can simply prepend the list when a new summary visualiser report is generated rather than rebuilding the index page on a separate scheduled workflow.

_includes/demo-list.md
mattyg: Moved from index.md -- treating all past runs as demo runs.

ThetaSinner:
> do we also want to stop uploading that data from PRs to influx?

That's less important to me. I don't think it's a bad thing that we're checking the integration of the summarizer and HTML visualiser. It's just producing permanent assets from those is what I'd like to avoid.

ThetaSinner (approved):

PR #415, opened 2025-12-17
refactor: remove countersigning scenario from CI workflows

matthme:
### Summary

Removes the countersigning scenario from `nomad.yaml` and `test.yaml` to not have to bother about flaky behaviour for the time being.

### TODO:

## Release Notes

* **Chores**
  * Removed testing for the "two_party_countersigning" scenario from CI workflows.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -183,14 +183,6 @@ jobs:
           echo
           df -h
 
-      - name: Smoke test - two_party_countersigning
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -183,14 +183,6 @@ jobs:
           echo
           df -h
 
-      - name: Smoke test - two_party_countersigning
```
matthme: Would commenting out be preferred rather than removing? I thought given that we likely won't bother about the scenario for quite a while removing is cleaner.

PR #423, opened 2025-12-18
fix: create telegraf metrics output dir if not found, to avoid the task failing and restarting 2m later

### LINKED ISSUE #375: Nomad task "report_host_metrics" fails to start on first execution
When first run, the nomad task `report_host_metrics` fails to start because its configured data directory `${NOMAD_ALLOC_DIR}/data/telegraf/metrics` does not exist.

Then it is scheduled to try running again in ~2 minute intervals

Once the `run_scenario` starts, it creates that data directory. On `report_host_metrics` next start, it will find the data directory and run sucessfully.

Because of the 2 minute restart schedule, it may be missing the host metrics from the first 2 minutes or less of the `run_scenario` task.

If we only want to collect host metrics while the `run_scenario` task is running, we could modify its configuration to retry running on a shorter interval. 

Alternatively we could have `report_host_metrics` create the data directory. Then we could either upload additional host metrics, or exclude metrics before the run_scenario task starts from being uploaded.

mattyg:
### Summary
Resolves #375 

I ran the Nomad CI workflow for 1 scenario and check its logs to confirm that `report_host_metrics` started successfully on the first try. Previously it would fail to start on the first try consistently.

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metrics startup reliability by ensuring required metrics storage is created before the monitoring agent starts, reducing failed or missed metric collections during initialization.

### DIFF
--- a/nomad/run_scenario.tpl.hcl
+++ b/nomad/run_scenario.tpl.hcl
@@ -77,8 +77,8 @@ job "{{ (ds "vars").scenario_name }}" {
           }
 
           config {
-            command = "telegraf"
-            args    = []
+            command = "sh"
+            args    = ["-c", "mkdir -p ${WT_METRICS_DIR} && telegraf --config ${NOMAD_TASK_DIR}/telegraf.host.conf"]
           }
         }
       }


nomad/run_scenario.tpl.hcl
```diff
@@ -77,8 +77,8 @@ job "{{ (ds "vars").scenario_name }}" {
           }
 
           config {
-            command = "telegraf"
-            args    = []
+            command = "sh"
+            args    = ["-c", "mkdir -p ${WT_METRICS_DIR} && telegraf --config ${NOMAD_TASK_DIR}/telegraf.host.conf"]
```
mattyg: Create directory if it doesn't exist, and pass in the config explicitly to avoid another error

veeso (approved):

ThetaSinner (approved):
Nice spot, and change looks convincing to me

PR #434, opened 2025-12-22
refactor: upload scenario bin to bucket, delete after scenario completes

### LINKED ISSUE #419: Ensure longer availability of scenario executables
Currently we are storing the compiled scenario executables as github artifacts.

In order to share the artifact URL to nomad clients, we use a [workaround](https://github.com/holochain/wind-tunnel/blob/d09bf8659b9b15aa09e15690a8935d919c096088/.github/workflows/nomad.yaml#L123) to get the final redirect url from the underlying storage provider with the authorization token included.

However that url is only valid for 5 minutes. If a nomad client has a slower internet connection, it may take them some time to get to the `run_scenario` task where the actual scenario is downloaded. If the 5 minutes has passed, then the artifact URL will return a 503.

See this allocation for an example: https://nomad-server-01.holochain.org:4646/ui/allocations/16e9a695-df73-984b-4557-2d423f69e4ec/run_scenario

The only officially supported way to download artifacts from another service is to use the github api, which requires authentication. We could grant all clients a github api token with limited read scope, so it could query to download the artifact, but that would not fit cleanly into nomad's "artifact" feature which expects to be given a URL.

I think the simplest solution would be to upload the scenario executables to an 3rd party storage bucket that is publically readable, and then delete from the bucket upon completion.

AC:
- Upload scenario executables to storage bucket (at the same provider used in https://github.com/holochain/wind-tunnel/pull/403 -- currently hetzner)
- Update scenario url passed to clients
- Delete scneario executable from storage bucket upon completion.

mattyg:
### Summary
Resolves #419 

Resolves https://github.com/holochain/wind-tunnel/issues/440

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Builds now produce ZIP archives of each build for distribution.
  * CI adds cleanup jobs to remove uploaded archived artifacts from cloud storage.

* **Changes**
  * Scenario uploads now go directly to cloud storage and produce downloadable URLs.
  * CLI wrapper renamed and now forwards all provided S3-style arguments.
  * CI workflow secrets updated; additional secrets added and some made required.

* **Documentation**
  * README updated with public artifact hosting and Nomad usage guidance.

### DIFF
--- a/.github/workflows/nomad-common.yaml
+++ b/.github/workflows/nomad-common.yaml
@@ -29,9 +29,11 @@ on:
       INFLUX_TOKEN:
         required: true
       HETZNER_HOLOCHAIN_INFRA_BUCKETS_ACCESS:
-        required: false
+        required: true
       HETZNER_HOLOCHAIN_INFRA_BUCKETS_SECRET:
-        required: false
+        required: true
+      CACHIX_AUTH_TOKEN:
+        required: true
 
 jobs:
   run-scenarios:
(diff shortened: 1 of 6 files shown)


.github/workflows/nomad.yaml
```diff
@@ -97,17 +97,23 @@ jobs:
...
-          name: ${{ env.JOB_NAME }}
-          if-no-files-found: error
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.HETZNER_HOLOCHAIN_INFRA_BUCKETS_ACCESS }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.HETZNER_HOLOCHAIN_INFRA_BUCKETS_SECRET }}
+          AWS_DEFAULT_REGION: hel1
+          AWS_ENDPOINT_URL: https://hel1.your-objectstorage.com
+          S3_BUCKET_NAME: wind-tunnel-public-artifacts
+        run: |-
+          OUT_PATH="${S3_BUCKET_NAME}/run-artifacts/${{ github.run_id }}/${{ matrix.scenario-name }}.zip"
```
mattyg: Upload scenario to directory named by github run id. Then when the run completes we can trash the whole directory.

.github/workflows/nomad-common.yaml
```diff
@@ -101,15 +103,20 @@ jobs:
...
-            ./result/bin/
-            ./result/happs/
-          name: ${{ matrix.job-name }}
-          if-no-files-found: error
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.HETZNER_HOLOCHAIN_INFRA_BUCKETS_ACCESS }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.HETZNER_HOLOCHAIN_INFRA_BUCKETS_SECRET }}
+          AWS_DEFAULT_REGION: hel1
+          AWS_ENDPOINT_URL: https://hel1.your-objectstorage.com
+          S3_BUCKET_NAME: wind-tunnel-public-artifacts
```
jost-s: I slightly prefer to define these once as a workflow variable instead of typing them out repeatedly.

jost-s (dismissed):

.github/workflows/nomad-common.yaml
```diff
@@ -374,6 +371,35 @@ jobs:
...
+        run: |-
+          nix run .#awscli-s3 rm "s3://${S3_BUCKET_NAME}/run-artifacts/${{ github.run_id }}" -- --recursive
+
+          echo "Deleted ${AWS_ENDPOINT_URL}/${S3_BUCKET_NAME}/run-artifacts/${{ github.run_id }}" >> "$GITHUB_STEP_SUMMARY"
```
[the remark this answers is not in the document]
mattyg: It works both ways but I agree your syntax is clearer.

cdunster (dismissed):
Looks good with one small suggestion.

cdunster (approved):

PR #447, opened 2026-01-07
fix: validation_receipts scenario gets stuck

### LINKED ISSUE #329: `validation_receipts` scenario gets stuck in infinite doesn't end

veeso:
### Summary

Refactored the scenario logic in order to prevent concurrent actions to not being validated due to creation failed.

closes #329

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added a validation_receipts scenario and included it in the CI run matrix to expand automated coverage.

* **Chores**
  * Reworked validation receipt tracking and completion monitoring for scenarios.
  * Updated workspace dependencies for the new scenario.
  * Added ignore rules for the Rust rover project.
  * Minor formatting and whitespace cleanups across scenario configuration files.

### DIFF
--- a/scenarios/validation_receipts/src/main.rs
+++ b/scenarios/validation_receipts/src/main.rs
@@ -101,11 +101,6 @@ fn agent_behaviour(
 
     get_receipts_for_action(ctx, wait_for_all)?;
 
-    // remove the action from pending if complete
-    if ctx.get().scenario_values.is_action_complete() {
-        ctx.get_mut().scenario_values.pending_action = None;
-    }
-
     Ok(())
 }
 


scenarios/validation_receipts/src/main.rs
```diff
@@ -61,84 +58,100 @@ fn agent_setup(
...
+    // keep checking until all receipts are complete
+    while !ctx.get().scenario_values.is_action_complete() {
+        let response: Vec<ValidationReceiptSet> = call_zome(
+            ctx,
+            "crud",
+            "get_sample_entry_validation_receipts",
+            action_hash.clone(),
+        )?;
+
+        if started_at.elapsed() > Duration::from_secs(60) {
```
mattyg: If we do want to keep the while loop in place, then this check should probably be before fetching the response or after processing it. Why fetch the response and wait for it if we're just going to drop it afterwards?

PR #451, opened 2026-01-08
feat: stop running agents when their behaviour returns an error

cdunster:
### Summary

When a behaviour returns an error, the Wind Tunnel agent currently just logs the error and repeatedly runs the behaviour. This means that if there is an unrecoverable error then the behaviour just constantly runs, fails, and spams the logs. This change makes it so that if the behaviour fails then the agent running it gets stopped, if the scenario wants to ignore certain errors then that should be handled by the behaviour itself.

### TODO:

## Release Notes

* **Bug Fixes**
  * Fixed agent error handling to properly mark execution state as incomplete and exit early when unexpected errors occur during agent behavior execution.

### DIFF
--- a/framework/runner/src/run.rs
+++ b/framework/runner/src/run.rs
@@ -220,6 +220,8 @@ pub fn run<RV: UserValuesConstraint, V: UserValuesConstraint>(
                                     log::error!(
                                         "Agent behaviour [{assigned_behaviour}] failed for agent {agent_name}: {e:?}"
                                     );
+                                    behaviour_ran_to_complete = false;
+                                    break;
                                 }
                             }
                         }


cdunster:
> Maybe we should provide an option instead.

I personally don't think that this should be an option because I think that that will just confuse the matter and mean that behaviours may behave differently depending on what that option is set to which I don't think is what we want.

cdunster:
> I don't really agree with this change. I think the scenario has to decide one way or the other about whether errors are fatal and now this will require scenarios to have similar logic to check for specific errors. That seems to me like moving the problem. If we are going to switch this default around, then:

To me, it doesn't seem clear that only one error, the `AgentBailError`, would cause the behaviour to error-out, and instead it would be more clear and expected if errors caused the behaviour to fail.

> * Have existing behaviors been checked?

I'm running the scenarios to check what happens with this change but from what I can see I don't know why we would want to continue the behaviour if we get an error unless we explicitly state so like in [the `write_validated_must_get_agent_activity` scenario](https://github.com/holochain/wind-tunnel/blob/4ed435e41e60f583f9cd31ee3261dcd5f658544f/scenarios/write_validated_must_get_agent_activity/src/main.rs#L130-L132).

> * Does it make sense to keep the `AgentBailError` error?

I was thinking of keeping it as a way of forcing the behaviour to fail for example checking a result of something and returning an error if it's not as desired.

Maybe I jumped the gun on opening this and instead I should create an issue for it to refine with the team.

ThetaSinner:
> it doesn't seem clear that only one error

That's a fair point, it's not obvious.

> it would be more clear and expected if errors caused the behaviour to fail

That was how it worked. This error handling logic wasn't in the first version. It seems fine until you get 30 minutes into running a scenario and a zome call returns an error for no good reason. Making the error cause an exit by default, has the same level of unexpected behavior to my mind.

> I'm running the scenarios to check what happens with this change

I think it requires a deeper look than that. Certainly for any scenarios that myself or neonphog worked on, we would have been expecting the current error logic.

> create an issue for it to refine with the team

We could do that, but we've started the conversation here now. I'd be happy to get on a call and talk this through so you can proceed.

---

What did you think of the suggestion to modify the wrapped Rust client to treat connection errors as fatal? That'd solve for the reason this came up I think. Though it might not cover all of what is wanted now.

cdunster:
After some discussion, closing in favour of #474.

Issue #427, opened 2025-12-18
Empty summarizer data from `local_signals`


mattyg:
The summarizer is failing due to no data `wt.custom.signal_batch_send` from the `local_signals` scenario.

See  https://github.com/holochain/wind-tunnel/actions/runs/20353839494/job/58485002725

PR #496, opened 2026-01-27
fix: ensure local_signals metrics written on early exit

lucksus:
Fixes https://github.com/holochain/wind-tunnel/issues/427

The local_signals scenario was failing in CI because metrics were only written at the end of the agent_behaviour function. If the scenario duration ended while waiting for signals (30s timeout), the function would be interrupted and metrics would never be written.

This fix implements a MetricsGuard using Rust's RAII pattern with Drop to ensure recv metrics are always written when the function exits, whether normally or through early termination. The send metric is now written immediately after the zome call completes.

Changes:
  - Move start timer to beginning of function
  - Create MetricsGuard that writes metrics in Drop implementation
  - Write send metric immediately after zome call
  - Recv metrics automatically written on function exit (normal or early)

Tested with 5/5 stress tests passing with 20s duration (previously failed 100% of the time).

### TODO:

## Summary by CodeRabbit

* **Bug Fixes**
  * Receive metrics are now reliably recorded even on early exit or unexpected termination.
  * Send metrics are emitted immediately after actions complete.
  * Final receive/send metrics are now emitted automatically at run end, replacing previous manual emission steps.

### DIFF
--- a/scenarios/local_signals/src/main.rs
+++ b/scenarios/local_signals/src/main.rs
@@ -39,7 +39,6 @@ fn agent_behaviour(
     let _metrics_guard = MetricsGuard {
         write_fn: Box::new({
             let reporter = reporter.clone();
-            let start = start.clone();
             let received_count = received_count.clone();
             move || {
                 let recv_elapsed_s = start.elapsed().as_secs_f64();


scenarios/local_signals/src/main.rs
```diff
@@ -32,10 +33,35 @@ fn agent_behaviour(
         }
     })?;
 
-    let start = Instant::now();
```
ThetaSinner: This change is semantic, not a simple code cleanup. The time taken to acquire the app client and set up the signal receiver shouldn't be measured.

scenarios/local_signals/src/main.rs
```diff
@@ -57,20 +83,23 @@ fn agent_behaviour(
...
+// Guard struct that writes recv metrics when dropped, ensuring they're always written even on early exit
+struct MetricsGuard {
+    write_fn: Box<dyn FnOnce() + Send>,
+}
 
-    Ok(())
+impl Drop for MetricsGuard {
+    fn drop(&mut self) {
+        // SAFETY: We use take_mut to safely extract the FnOnce from the Option
+        // This is safe because Drop is only called once
```
ThetaSinner: Comment does not match the code

scenarios/local_signals/src/main.rs
```diff
@@ -32,10 +33,35 @@ fn agent_behaviour(
...
+                    ReportMetric::new("signal_batch_recv").with_field("value", recv_elapsed_s);
+                reporter.clone().add_custom(metric);
+
+                let metric = ReportMetric::new("signal_success_ratio")
+                    .with_field("value", count as f32 / 10_000.0);
+                reporter.clone().add_custom(metric);
+            }
+        }),
+    };
+
```
ThetaSinner: I think defining start probably belongs here?

PR #506, opened 2026-02-02
fix: `write_validated_must_get_agent_activity` scenario failed to produce chain_len metrics

### LINKED ISSUE #431: Summarizer can fail on getting `write_validated_must_get_agent_activity_chain_len` after scenario doesn't produce any values for that metric
https://github.com/holochain/wind-tunnel/actions/runs/20378593827/job/58568194485?pr=429

The summarizer fails with the following output:

```
[2025-12-19T19:23:36Z DEBUG holochain_summariser::query] Querying: ReadQuery { queries: ["SELECT value, zome_name, fn_name FROM \"windtunnel\".\"autogen\".\"wt.instruments.operation_duration\" WHERE run_id = 'zero_arc_create_and_read_20378593827' AND (operation_id = 'app_call_zome' OR operation_id = 'trycp_app_call_zome') AND is_error = 'true'"] }
Error: 1 out of 19 summaries failed:
[
    Error {
        context: "Write validated must_get_agent_activity summary",
        source: Error {
            context: "Load write_validated_must_get_agent_activity_chain_len data",
            source: NoSeriesInResult {
                table: "wt.custom.write_validated_must_get_agent_activity_chain_len",
                result: Object {
                    "statement_id": Number(0),
                },
            },
        },
    },
]
```

The question then is why no values were produced for that metric. It suggests that no `must_get_agent_activity` calls succeeded during the time that the scenario was running.

It is valid for the summarizer to report the issue. However, it's unhelpful to have the scenario not appear stable on CI. The suggestion for this issue is a brief assessment of the scenario:
- [ ] Do the setup and agent setup hooks properly prepare the scenario to do useful work?
- [ ] Does the scenario need more run time on CI than it is given in order to reliably start doing useful work?
- [ ] Are there any obvious issues with how the metric in question is reported that could have led to the failure?

veeso:
### Problem

The `write_validated_must_get_agent_activity` scenario was failing to produce `write_validated_must_get_agent_activity_chain_len` metrics. The root cause was that the
`get_random_agent_with_write_behaviour` zome call was timing out (using the default websocket timeout), causing the agent behaviour to fail before reaching the code that reports metrics.

### Solution

Changed the `get_random_agent_with_write_behaviour` to use a 15 seconds timeout.

Exposed the `call_zome_with_options` functions from the sdk

Closes #431

## Issue analysis

The issue is related to failures at the startup of the scenario, where websocket timeouts with this:

```log
[2026-01-30T11:02:49Z ERROR wind_tunnel_runner::run] Agent behaviour [must_get_agent_activity] failed for agent agent-0: Conductor API error: ExternalApiWireError(InternalError("Wasm runtime error while working with Ribosome: RuntimeError: WasmError { file: \"crates/holochain/src/core/ribosome/host_fn/get_links.rs\", line: 76, error: Host(\"tx5 send error to peer at url wss://dev-test-bootstrap2.holochain.org:443/kHk5tq7ziSbmRbw2yrSVXHCZe3Ajv_PMj1bHdKxM-kM (src: Peer connection failed)\") }"))
[2026-01-30T11:03:49Z ERROR wind_tunnel_runner::run] Agent behaviour [must_get_agent_activity] failed for agent agent-0: Conductor API error: WebsocketError(Timeout(Elapsed(())))
2026-01-30T11:13:05.335241Z ERROR holochain::core::ribosome::real_ribosome: crates/holochain/src/core/ribosome/real_ribosome.rs:652: runtime_error=RuntimeError { source: Sys(User(WasmError { file: "crates/holochain/src/core/ribosome/host_fn/get_links.rs", line: 76, error: Host("tx5 send error to peer at url wss://dev-test-bootstrap2.holochain.org:443/Z8eeyujuvaI3hrCCXebmyI3zFiH5izqG3h3PlaZuRRo (src: timed out)") })), wasm_trace: [] } zome=Zome { name: ZomeName("validated_must_get_agent_activity"), def: Wasm(WasmZome { wasm_hash: WasmHash(uhCok5GxPE4BzdWTadkgvC5iXzLUKJ2AdFslMv1htxfev0-EYyCRS), dependencies: [ZomeName("validated_must_get_agent_activity_integrity")] }) } fn_name=FunctionName("get_random_agent_with_write_behaviour")
```

This error can happen for several time, on average just twice, but could happen more times. Every time it happens it block the behaviour for like one minute.

After that, it start querying for the random write peer agent, which usually takes 3 minutes to complete, but it's not deterministic. Sometimes it takes seconds, sometimes it takes 7 minutes.

In some cases, it can even cause the behaviour to fail again with this issue:

```log
2026-01-30T11:19:06.665280Z ERROR kitsune2_gossip::gossip: /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/kitsune2_gossip-0.3.2/src/gossip.rs:361: Peer behavior error: PeerBehaviorError { ctx: "Session id mismatch: b\" \\xa0\\xee\\x9e\\x9d\\xfc\\xfa\\x83\\xae8\\xfc\\xdd\" != b\"\\xfe\\xec\\x90t\\x88\\x1a\\x859\\x12\\xb5\\xf1K\"" }
```

In some rare case it never succeds to get the random peer agent due to too many ws timeouts. Reducing the ws timeout is enough to have enough time to succeed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced agent activity batch processing with improved error recovery and handling
  * Strengthened logging and monitoring across agent workflows
* **Chores**
  * Updated project dependencies

### DIFF
--- a/bindings/client/src/app_websocket.rs
+++ b/bindings/client/src/app_websocket.rs
@@ -2,7 +2,7 @@ use crate::ToSocketAddr;
 use crate::error::handle_api_err;
 use anyhow::Result;
 use holo_hash::DnaHash;
-use holochain_client::{AgentSigner, AppWebsocket, ZomeCallTarget};
+use holochain_client::{AgentSigner, AppWebsocket, CallZomeOptions, ZomeCallTarget};
 use holochain_conductor_api::{AppAuthenticationToken, AppInfo};
 use holochain_types::app::{DisableCloneCellPayload, EnableCloneCellPayload};
 use holochain_types::prelude::{
@@ -66,9 +66,10 @@ impl AppWebsocketInstrumented {
         zome_name: impl Into<ZomeName> + Clone,
         fn_name: impl Into<FunctionName> + Clone,
         payload: ExternIO,
+        options: CallZomeOptions,
     ) -> anyhow::Result<ExternIO> {
         self.inner
-            .call_zome(target, zome_name.into(), fn_name.into(), payload)
+            .call_zome_with_options(target, zome_name.into(), fn_name.into(), payload, options)
             .await
             .map_err(handle_api_err)
     }
@@ -130,6 +131,7 @@ fn pre_call_zome(
     zome_name: &(impl Into<ZomeName> + Clone),
     fn_name: &(impl Into<FunctionName> + Clone),
     _payload: &ExternIO,
+    _options: &CallZomeOptions,
 ) {
     let zome_name: ZomeName = zome_name.clone().into();
     let fn_name: FunctionName = fn_name.clone().into();
--- a/scenarios/write_validated_must_get_agent_activity/src/main.rs
+++ b/scenarios/write_validated_must_get_agent_activity/src/main.rs
@@ -24,13 +24,17 @@ impl UserValuesConstraint for ScenarioValues {}
 fn agent_setup(
     ctx: &mut AgentContext<HolochainRunnerContext, HolochainAgentContext<ScenarioValues>>,
 ) -> HookResult {
+    log::info!("Starting conductor and installing app...");
     start_conductor_and_configure_urls(ctx)?;
+    log::info!("Conductor started.");
     install_app(
         ctx,
         happ_path!("validated_must_get_agent_activity"),
         &"validated_must_get_agent_activity".to_string(),
     )?;
+    log::info!("App installed.");
     try_wait_for_min_agents(ctx, Duration::from_secs(120))?;
+    log::info!("Minimum number of agents reached.");
 
     // 'write' peers create a link to announce their behaviour so 'get_agent_activity' peers can find them
     if ctx.assigned_behaviour() == "write" {
@@ -41,6 +45,10 @@ fn agent_setup(
             (),
         )?;
     }
+    log::info!(
+        "Agent setup complete for behaviour '{}'.",
+        ctx.assigned_behaviour()
+    );
 
     Ok(())
 }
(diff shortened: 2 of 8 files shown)


bindings/runner/src/holochain_runner.rs
```diff
@@ -175,11 +175,15 @@ impl HolochainRunner {
         .context("Failed to write conductor config file")?
         .to_path_buf();
 
+        // get WT_HC_LOG env var and pass it to the child process
+        let wt_hc_log = std::env::var("WT_HC_LOG").unwrap_or_else(|_| "error".to_string());
```
mattyg:
```suggestion
        let wt_hc_log = std::env::var("WT_HOLOCHAIN_RUST_LOG").unwrap_or_else(|_| "error".to_string());
```

If we do keep, nitpick to request clearer name.

scenarios/write_validated_must_get_agent_activity/src/main.rs
```diff
@@ -79,151 +88,172 @@ fn agent_behaviour_must_get_agent_activity(
...
-                            .scenario_values
-                            .currently_pending_batch_chain_top = Some(batch_chain_top.clone());
-                        batch_chain_top
-                    }
-                    Err(_) => {
-                        // The next batch may not yet have been created by the writing peer.
-                        return Ok(());
-                    }
-                }
+    // get write peer if we don't have one yet
```
mattyg: Are there any functional changes here? Or just making the code more readable?

mattyg:
Just to clarify -- have you confirmed this resolves the issue, or you haven't been able to reproduce it reliably enough do so?

bindings/runner/src/holochain_runner.rs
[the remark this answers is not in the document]
veeso: That's exactly what I want to avoid. If you enable debug it's basically impossible to debug the scenario, unless you write to file, which is annoying. Holochain is just too verbose, and so it's basically impossible to catch the scenario logs.

In this investigation it would have been impossible for me to understand what was going on without this flag. I needed the debug level on the scenario, but hc was just making noise.

veeso:
> Just to clarify -- have you confirmed this resolves the issue, or you haven't been able to reproduce it reliably enough do so?

I was able to reproduce it, even if it was very unlikely, and it was due to too many ws timeouts.

scenarios/write_validated_must_get_agent_activity/src/main.rs
```diff
@@ -79,151 +88,172 @@ fn agent_behaviour_must_get_agent_activity(
...
+    );
+
     std::thread::sleep(Duration::from_secs(1));
 
     Ok(())
 }
 
 fn main() -> WindTunnelResult<()> {
+    let mut websocket_config = WebsocketConfig::CLIENT_DEFAULT;
+    websocket_config.default_request_timeout = Duration::from_secs(15);
```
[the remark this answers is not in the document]
veeso: I've added this possibility by providing ids and configurations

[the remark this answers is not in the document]
ThetaSinner: I'm not tracking the changes in this file really yet, I need to review again tomorrow when I'm fresh. The change to add `call_zome_with_client` in the common code is fair enough I think. The rest of the logic for having multiple app clients, I think that shouldn't go in the common code at all. It's fine to connect a custom client within a scenario if it's needed and just put it in the scenario values. The common stuff is a convenience where things are expected to be often needed across scenarios.

I will get my head around why we need a lower timeout soon - though on first look that seems like the scenario is doing what it should by telling us that something is slow - but the other option for solving this is to modify the Rust client to support overriding the timeout when calling a zome function. Not as nice in a statically typed language as a dynamic one, but we have that option in the JavaScript client.

bindings/runner/src/holochain_runner.rs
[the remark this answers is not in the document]
mattyg: Can't you do that via the RUST_LOG var? i.e. `RUST_LOG=debug,holochain=off`?

veeso:
@ThetaSinner I get your point, but atm I don't see any workaround.

The holochain client lib doesn't support to change the timeout.

We can't create a new client inside of the scenario as well.

So we have two options:

- Allow the user to create clients with different configs as-is
- Add timeout call to holochain client lib.

veeso:
@ThetaSinner @mattyg I've upgraded hc client with the call options, so now I've reverted all the changes and just used call_zome_with_options by passing 15 seconds to that single call.

bindings/client/src/app_websocket.rs
```diff
@@ -73,6 +73,21 @@ impl AppWebsocketInstrumented {
             .map_err(handle_api_err)
     }
 
+    #[wind_tunnel_instrument(prefix = "app_", pre_hook = pre_call_zome_with_options)]
+    pub async fn call_zome_with_options(
```
ThetaSinner: The function name gets picked up and used in the metrics - this will cause a funky change to the summarizer results. I wonder if extending the macro with a "rename" option might be good?

bindings/client/src/app_websocket.rs
```diff
@@ -73,6 +73,21 @@ impl AppWebsocketInstrumented {
...
+        &self,
+        target: ZomeCallTarget,
+        zome_name: impl Into<ZomeName> + Clone,
+        fn_name: impl Into<FunctionName> + Clone,
+        payload: ExternIO,
+        options: CallZomeOptions,
+    ) -> anyhow::Result<ExternIO> {
+        self.inner
+            .call_zome_with_options(target, zome_name.into(), fn_name.into(), payload, options)
```
jost-s: Maybe the above call `call_zome` could also call this `call_zome_with_options`, with the default for `CallZomeOptions`? Same inside of the `AppWebSocket` actually, the plain `call_zome` could call the options method with defaults for `CallZomeOptions`.

bindings/client/src/app_websocket.rs
veeso: I've simplified by just using call_zome but with the option argument. The only point it was  used is `common.rs`, where I'm now using `call_zome_with_options` with the default options.

PR #577, opened 2026-03-16
feat(nomad): support arbitrary env vars in scenario vars.json files (#540)

### LINKED ISSUE #540: Deploy scenario with configuration
Some scenarios can be configured with env variables. However we don't have a way to apply configuration to nomad deployments.

One option could be modify the nomad job vars.json file to include an object of arbitrary key values, which would then be set as env variables when the scenario is run.

AC:

* Specify arbitrary key-value pairs in the vars.json file (applied to all nodes), in the following format:

  ```
    "env": {
      "MY_VARIABLE": "value"
    }
  ```
* Read from file and apply to env block in run_scenario nomad template
* Refactor custom way of setting MIN_AGENTS env variable, instead using this same arbitrary key-value setting
* Update existing scenario vars files for existing scenarios with env configuration options (unyt, validation receipts, get agent activity volatile), set to default values for those options

veeso:
### Summary

Replace the hardcoded MIN_AGENTS template variable with a generic env
object in vars.json files. Each key-value pair in the env object is
injected into the Nomad task's env block via a gomplate range loop.

- Add env object support to gomplate template (run_scenario.tpl.hcl)
- Migrate unyt_chain_transaction: move min_agents to env.MIN_AGENTS,
  add NUMBER_OF_LINKS_TO_PROCESS with default value
- Add env vars to write_get_agent_activity_volatile (CONDUCTOR_* vars)
- Add env object to validation_receipts (empty, supports NO_VALIDATION_COMPLETE)
- Add env vars to remote_signals (SIGNAL_INTERVAL_MS, RESPONSE_TIMEOUT_MS)
- Update README documentation with env object syntax and example

Closes #540

### TODO:

## Summary by CodeRabbit

* **Documentation**
  * Updated configuration syntax to use environment variables for scenario-specific settings, replacing the previous min_agents parameter with a generalized env object.

* **Chores**
  * Migrated job configuration from explicit parameters to environment variable-based configuration across all job variants.

### DIFF
--- a/nomad/job-variants/canonical/vars/unyt_chain_transaction_zero_arc.json
+++ b/nomad/job-variants/canonical/vars/unyt_chain_transaction_zero_arc.json
@@ -1,7 +1,6 @@
 {
   "scenario_name": "unyt_chain_transaction_zero_arc",
   "duration": 900,
-  "min_agents": 7,
   "assignments": [
     {
       "behaviour": "initiate"
@@ -20,5 +19,9 @@
     {
       "behaviour": "zero_observer"
     }
-  ]
+  ],
+  "env": {
+    "MIN_AGENTS": "7",
+    "UNYT_NUMBER_OF_LINKS_TO_PROCESS": "10"
+  }
 }
--- a/nomad/README.md
+++ b/nomad/README.md
@@ -24,6 +24,31 @@ A simple Example:
 }
 ```
 
+An example with environment variables:
+
+```json
+{
+  "scenario_name": "write_get_agent_activity_volatile",
+  "duration": 900,
+  "assignments": [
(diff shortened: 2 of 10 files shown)


nomad/job-variants/canonical/vars/unyt_chain_transaction_zero_arc.json
```diff
@@ -20,5 +19,8 @@
     {
       "behaviour": "zero_observer"
     }
-  ]
+  ],
+  "env": {
```
cdunster: Should `UNYT_NUMBER_OF_LINKS_TO_PROCESS` not be set here too?

PR #603, opened 2026-03-27
feat: include zome calls made in unyt UI in transactor loop of both unyt wind tunnel scenarios

### LINKED ISSUE #563: Add missing zome fn calls to existing unyt scenarios
Add the additional calls in the unyt scenario required to evaluate if we are meeting their performance requirements. This is a follow-up to [https://github.com/holochain/wind-tunnel/issues/463#issuecomment-3975552700](<https://github.com/holochain/wind-tunnel/issues/463#issuecomment-3975552700>)

The following zome fns are not called in the wind tunnel scenario at all. They could be included in the existing scenario behavior:

* whoami
* get_smart_agreement
* get_all_notification_links
* get_history -- Currently only called on teardown. Example of usage in unyt sweettests: [https://github.com/unytco/unyt-app/blob/develop/sweettest/tests/transaction/history_pagination.rs](<https://github.com/unytco/unyt-app/blob/develop/sweettest/tests/transaction/history_pagination.rs>))
* check_agent_exists
* get_global_units_details

The Unyt UI calls these 4 zome fns simultaneously when the Actions list is refreshed, We should reproduce in the wind tunnel scenario, and record the overall time that all complete as a single metric.

* get_requests_to_execute_agreements
* get_actionable_transactions (Currently called individually)
* get_incoming_raves (Currently called individually)
* get_sorted_requests_to_spend

The Unyt UI calls these 3 zome fns sequentially. We should reproduce in the wind tunnel scenario. We should record the time that all complete as a single metric.

* get_transaction
* get_status
* get_transaction on related_transaction

AC:

* Add these to the existing unyt scenario, in the transactor loop
* Record a metric for the duration it takes to complete *all* of these calls, so that we can distinguish from the normal transactor times.

mattyg:
### Summary
Resolves #563 

### TODO:

## Summary by CodeRabbit

* **New Features**
  * UI-triggered refresh now updates identity, agreements, notification links, agent availability, global unit details, and actionable items; per-transaction detail refreshes surface primary/related transaction status with aggregated metrics.
  * Action list and related data fetches run in parallel for faster results and graceful fallbacks; history is fetched earlier for timelier activity.

* **Bug Fixes / Reliability**
  * Improved timing, metrics, and error tracking for refresh stages and remote calls.

* **Chores**
  * Added a raw zome-call helper that records and reports operation outcomes.


jost-s (changes requested):
Overall looks good. The metrics must be added individually.

scenarios/unyt_chain_transaction_zero_arc/README.md
```diff
@@ -82,6 +82,25 @@ The scenario records several custom metrics:
 - `wt.custom.completed_transaction_spends`: Records the count of completed spend transactions at scenario teardown
```
veeso, line 82: Update `unyt_propose` metrics

veeso (changes requested):
Must also include metrics for `unyt_propose` scenario

scenarios/unyt_chain_transaction_zero_arc/README.md
jost-s: @veeso Please explain in full what you mean. This is a PR that adds zome call metrics to the `spend` behavior of the `unyt_chain_transaction` scenarios.

bindings/runner/src/common.rs
```diff
@@ -740,6 +741,45 @@ where
...
+            )
+            .await?;
+
+        result
+            .decode()
+            .map_err(|e| anyhow::anyhow!("Decoding failure: {e:?}"))
+    }
+    .await;
+
+    report_operation(reporter, operation_record, &result);
```
cdunster: The existing `call_zome_with_options` doesn't call `report_operation`, is that a problem? Could `call_zome_with_options` now just become a wrapper around this `call_zome_raw` function?

scenarios_common/unyt_scenario/src/behaviour/spend.rs
```diff
@@ -47,20 +60,246 @@ pub fn agent_behaviour<SV: UnytScenarioValues>(
...
+            Ok(refresh_results) => {
+                let failed_calls = [
+                    refresh_results.actionable_transactions.is_err(),
+                    refresh_results.incoming_raves.is_err(),
+                    refresh_results.requests_to_execute_agreements.is_err(),
+                    refresh_results.sorted_requests_to_spend.is_err(),
+                ]
+                .into_iter()
+                .filter(|is_error| *is_error)
+                .count() as u64;
```
cdunster: I like how clean this is but it's basically the same as the if-statements below and by keeping it separate I'm worried that if we add more errors to count in the future then we'll add the count but forget the log or the other way around. Therefor, maybe it's better to just do:

```rust
let mut failed_calls = 0_u64;
if let Err(err) = &refresh_results.actionable_transactions {
    failed_calls = failed_calls.saturating_add(1);
    log::warn!(
        "get_actionable_transactions failed during action list refresh: {err}"
    );
}
```

scenarios_common/unyt_scenario/src/behaviour/spend.rs
```diff
@@ -47,20 +60,246 @@ pub fn agent_behaviour<SV: UnytScenarioValues>(
...
+            ReportMetric::new("ui_transaction_detail_refresh_related_transaction_failed_calls")
+                .with_field("value", detail_outcome.related_transaction_failed_calls),
+        );
+    }
+
+    // Refresh history list
+    match ctx.unyt_get_history(Pagination {
+        high_boundary: None,
+        per_page: 10,
+    }) {
```
cdunster, line 262: Unless something in Unyt has changed, this will only ever get a maximum of 10 items, is that okay? To get the entire history you need to loop until `history.end_of_chain` is `true`. Something like:
```rust
    while let Ok(History {
        items,
        low_boundary,
        end_of_chain,
    }) = ctx.unyt_get_history(Pagination {
        high_boundary: current_boundary,
        per_page: 10,
    }) {
        current_boundary = Some(low_boundary);

        // Do something with items here.

        if end_of_chain {
            break;
        }
    }
```

scenarios/unyt_chain_transaction_zero_arc/README.md
veeso: Ah okay, I thought it also added to other behaviours

veeso (dismissed):

scenarios_common/unyt_scenario/src/behaviour/spend.rs
jost-s: I believe that the UI only queries the last 10 items, because on the dashboard not all items of the history are displayed. There are recent items and a watch list. But I'll double-check with the Unyt team.

jost-s: Joel confirms the UI is just fetching the most recent 10 items.

bindings/runner/src/common.rs
jost-s: I don't know if it's a problem, but I think it's fine to always report zome calls.

@ThetaSinner What's your thinking?

scenarios_common/unyt_scenario/src/behaviour/spend.rs
jost-s: I agree https://github.com/holochain/wind-tunnel/pull/603/commits/9b708c66772e81f6779ff8f9dedc228be1300a83

bindings/runner/src/common.rs
ThetaSinner: It's already the instrumented client, is this not already a duplicate report here?

jost-s: That's right. This is a duplication. I'll revert it.

jost-s: https://github.com/holochain/wind-tunnel/pull/603/changes/3abc756c95a3c908d1457206330345c996210548

jost-s: I reverted the other duplicate reporting as well: https://github.com/holochain/wind-tunnel/pull/603/changes/23e695859843d44cd7659af5f42ba21b50ed8062

All unyt zome calls were reporting on top of the already instrumented app client.

scenarios_common/unyt_scenario/src/behaviour/spend.rs
```diff
@@ -47,20 +60,241 @@ pub fn agent_behaviour<SV: UnytScenarioValues>(
...
+                reporter.add_custom(
+                    ReportMetric::new("ui_action_list_refresh_failed_calls")
+                        .with_field("value", failed_calls),
+                );
+            }
+            Err(err) => {
+                log::warn!("Action list grouped refresh failed: {err}");
+                reporter.add_custom(
+                    ReportMetric::new("ui_action_list_refresh_failed_calls")
+                        .with_field("value", 4_u64),
```
cdunster: I guess this is `4` because if the call fails then it counts as all 4 things above failing? This seems like an accident waiting to happen though, I'm not sure how to change it other than maybe creating a separate metric but if we count more failures in the future then we might forget to increment this.

jost-s: This was totally worth it. The return value didn't need to be a `Result`. That eliminates the need for the `Err` arm.

https://github.com/holochain/wind-tunnel/pull/603/commits/3e932d9372122452a613908194c44e420beaee2d

cdunster (dismissed):

cdunster (approved):

jost-s (approved):

PR #607, opened 2026-04-01
test: fix `say_something_to_other_chatter` test

### LINKED ISSUE #579: fix `say_something_to_other_chatter` test hanging on ci
This test `say_something_to_other_chatter` is hanging on macOS CI.

## Main hypothesis: missing kitsune cleanup causes tokio runtime to hang

The test never performs graceful shutdown of kitsune resources. When the test function returns, `WtChatter` instances are dropped but the kitsune background tokio tasks (gossip, bootstrap polling, transport connections) keep running because they never receive a shutdown signal. The `#[tokio::test]` runtime waits for all spawned tasks to complete before exiting, so the test hangs indefinitely.

### Evidence

The kitsune2 own `shutdown_space` integration test (`kitsune2-0.4.0-dev.4/tests/integration.rs:328`) demonstrates the correct cleanup sequence:

```rust
// 1. Leave the spaces (signals tasks to wind down)
space.local_agent_leave(agent).await;
// 2. Wait for peer tombstones
// 3. Drop space references
drop(space);
// 4. Remove spaces from kitsune
kitsune.remove_space(SPACE_ID).await.unwrap();
// 5. Wait for task count to return to initial
```

Our test does **none** of these steps — it just returns, dropping everything without signaling shutdown.

A similar issue was observed in the messaging test on the Wind Tunnel framework, where tokio tasks were kept alive after shutdown causing indefinite waits.

### Proposed fix

Add explicit cleanup to the test before returning: call `local_agent_leave` for both agents, then `remove_space` on the kitsune instances. Consider adding a `WtChatter::shutdown()` method to encapsulate this for reuse.

## Alternative checks if the main hypothesis doesn't resolve it

1. **Timeouts on `WtChatter::create`**: The `kitsune_builder.build().await` and `kitsune.space(...).await` calls have no timeout. If Iroh transport setup stalls on macOS CI (e.g. due to network restrictions), this would hang before the test even starts. Wrap these in `tokio::time::timeout`.

2. **Relay URL misconfiguration**: The test reuses the bootstrap server address as the Iroh relay URL (`let relay_url = format!("http://{}", bootstrap_server.listen_addrs()[0])`). This is not a real relay server. On macOS, the Iroh transport may handle an unreachable/wrong relay differently than Linux — hanging instead of erroring. Verify whether this is the intended setup.

3. **macOS CI network restrictions**: macOS CI runners may have stricter network sandboxing (restricted UDP, mDNS). The Iroh transport may stall on peer connection attempts that would succeed on Linux. Check if the test needs macOS-specific transport configuration or should be skipped on macOS.

4. Check all iroh shutdown sequence inside of kitsune tests

## AC

- [ ] The test is no more ignored and is passing on ci.

veeso:
### Summary

updating to kitsune 0.4.0-dev.7

closes #579

### TODO:

## Summary by CodeRabbit

* **New Features**
  * Added an explicit shutdown action to cleanly release networking and session resources.

* **Bug Fixes**
  * Improved test reliability with explicit cleanup, forced runtime shutdown, better timeout propagation, and added debug logging.

* **Refactor**
  * Simplified state and resource management for more predictable connection lifecycle and shutdown behavior.

* **Tests**
  * CI test runner now emits test output and increases logging verbosity for easier debugging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,41 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
+dependencies = [
+ "crypto-common 0.1.7",
+ "generic-array",
+]
+
+[[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures 0.2.17",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.10.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
+]
+
+[[package]]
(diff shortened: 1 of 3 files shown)


ThetaSinner (commented):
I like the fix, great to make use of Kitsune2's own shutdown mechanism. I think this PR works as it is but I left some very minor comments and agreed with the rabbit.

ThetaSinner (approved):

PR #618, opened 2026-04-16
Publish scenarios to releases and pre-releases

### LINKED ISSUE #608: Only build scenarios on a change and push to a release
Currently, the scenarios are re-built on every run of the CI workflows. This is unnecessary as the binaries themselves don't change that often, just new scenarios are added.

The idea of this issue is to build all the scenarios on a release and then push them as artefacts to that release.

If the scenarios do change as part of a PR then they do need to be rebuilt and the new binaries need to be tested. If possible, it would be nice to only rebuild the affected scenarios and use the prebuilt binaries of the unaffected scenarios.

AC:
- [ ] Upon a release, all the scenarios are built and uploaded as release artefacts
- [ ] Upon a push to `main`, create a new pre-release with the new scenario binary ZIP files - should be some cleanup around old pre-releases.
- [ ] In CI, if any of the scenario, Wind Tunnel, or Zome code changes then the scenarios are rebuilt and the new scenarios are used to test.
  - (_Stretch goal_) Only the affected scenarios are rebuilt

cdunster:
### Summary

This PR is part of #608, it automatically creates a pre-release with the name `main-<commit-sha>` on a push to any of the main branches. There is also a new workflow that builds and publishes the scenario zip files whenever a new release (or pre-release) is published, this workflow can also be triggered manually to override existing scenario zips or add them to old releases.

### TODO:

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -7,8 +7,50 @@ on:
       - main-*
 
 jobs:
-  call:
+  full_release:
     uses: holochain/actions/.github/workflows/publish-release.yml@v1.6.0
     secrets:
       HRA2_GITHUB_TOKEN: ${{ secrets.HRA2_GITHUB_TOKEN }}
       HRA2_CRATES_IO_TOKEN: ${{ secrets.HRA2_CRATES_IO_TOKEN }}
+
+  pre_releases:
+    name: Manage pre-releases
+    needs: [full_release]
+    if: github.ref_name == 'main'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-tags: true
+
+      - name: Delete old pre-releases
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release list --json tagName,isPrerelease \
+            --jq '.[] | select(.isPrerelease and (.tagName | startswith("main-"))) | .tagName' \
+            | xargs -r -I % gh release delete % --yes --cleanup-tag
+
+      - name: Check if a full release exists for this commit
+        id: check_release
+        run: |
+          if [[ -n $(git tag --points-at HEAD) ]]; then
+            echo "A tag is on this commit, skipping pre-release creation."
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Create pre-release
+        if: steps.check_release.outputs.exists == 'false'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REF_NAME: ${{ github.ref_name }}
+        run: |-
+          tag="main-$(git rev-parse --short HEAD)"
+          gh release create "$tag" \
+            --prerelease \
+            --title "Scenarios pre-release ($tag)" \
+            --notes "Scenario binaries built from commit $(git rev-parse HEAD) on branch ${REF_NAME}."
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -18,6 +18,9 @@ jobs:
     needs: [full_release]
     if: github.ref_name == 'main'
     runs-on: ubuntu-latest
+    concurrency:
+      group: pre-release
+      cancel-in-progress: true
     permissions:
       contents: write
     steps:
--- a/.github/workflows/publish-scenarios.yml
+++ b/.github/workflows/publish-scenarios.yml
@@ -1,0 +1,59 @@
+name: Publish scenarios
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish_scenarios:
+    name: Build and publish the scenarios
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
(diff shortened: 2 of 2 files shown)


.github/workflows/publish-scenarios.yml
```diff
@@ -0,0 +1,67 @@
...
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to publish scenarios for"
+        required: true
+        type: string
```
cdunster, line 11: This manual trigger allows triggering this build and publishing of scenarios to existing scenarios and allows us to re-trigger the build if we need to.

.github/workflows/publish-scenarios.yml
```diff
@@ -0,0 +1,67 @@
+name: Publish scenarios
+
+on:
+  release:
+    types: [published]
```
cdunster, line 5: AFAIK, this should only trigger on the first publish of a release (not drafts though)

.github/workflows/publish-release.yml
```diff
@@ -12,3 +12,49 @@ jobs:
...
+          fi
+
+      - name: Delete old pre-releases
+        if: steps.check_release.outputs.skip == 'false'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release list --json tagName,isPrerelease \
+            --jq '.[] | select(.isPrerelease and (.tagName | startswith("main-"))) | .tagName' \
+            | xargs -r -I % gh release delete % --yes --cleanup-tag
```
cdunster, line 37: Immediately deletes the old pre-release upon a new push to one of the main branches. This could cause issues if a job is currently running. I'm open to other suggestions about when to clean-up the temporary pre-releases, maybe when a full release is created?

.github/workflows/publish-release.yml
```diff
@@ -12,3 +12,49 @@ jobs:
...
+
+      - name: Create pre-release
+        if: steps.check_release.outputs.skip == 'false'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REF_NAME: ${{ github.ref_name }}
+        run: |-
+          tag="main-$(git rev-parse --short HEAD)"
```
cdunster, line 55: The created release and tag will have the name like `main-<sha>`, I think that this tag will be ignored by our changelog generation but that needs to be checked when we run this for the first time.

.github/workflows/publish-release.yml
cdunster: From Claude on the follow-up PR (#613) about this:

> **One issue found.**
> 
> **Race condition between pre-release creation and scenario upload** — `.github/actions/nomad/action.yaml` lines 139–150, `.github/workflows/publish-release.yml` lines 41–60.
> 
> `publish-release.yml` deletes old pre-releases and immediately creates a new empty one. This fires the `release: published` event, which triggers `publish-scenarios.yml` to build and upload scenarios — but nix builds take many minutes. During that window, `get-scenario-url` calls `gh release list --limit 1`, gets the new empty pre-release tag, and constructs a URL for a zip that doesn't exist yet. Any nomad run in this window (including the scheduled ones) will fail at download time with a 404.
> 
> The simplest fix is to move the `Delete old pre-releases` step into `publish-scenarios.yml`, running it just before `Upload scenarios to GitHub release`. That way the old pre-release stays valid until new scenarios are ready, and `--clobber` on the upload handles any tag reuse.
> 
> ```yaml
> # in publish-scenarios.yml, before "Upload scenarios to GitHub release":
> - name: Delete old pre-releases
>   env:
>     GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
>   run: |
>     gh release list --json tagName,isPrerelease \
>       --jq '.[] | select(.isPrerelease and (.tagName | startswith("main-"))) | .tagName' \
>       | grep -v "^${{ inputs.tag || github.event.release.tag_name }}$" \
>       | xargs -r -I % gh release delete % --yes --cleanup-tag
> ```

cdunster: This has been somewhat addressed by searching for the scenario within the latest 10 releases instead of just the latest release. It does mean that if a new release is created during a run then it is possible that the first bunch of runs will use the old scenarios from the latest release at the time and then rest will use the scenarios from the new release that was created during the run.

ThetaSinner: Can we put this workflow in a group so they're forced to run in serial?

ThetaSinner (dismissed):

ThetaSinner (approved):

.github/workflows/publish-scenarios.yml
```diff
@@ -0,0 +1,73 @@
...
+jobs:
+  publish_scenarios:
+    name: Build and publish the scenarios
+    runs-on: ubuntu-latest
+    concurrency:
+      # Pre-release builds share a cancellable group with publish-release.yml
+      # so a new push to main cancels an in-progress build. Full release builds
+      # use the tag name as the group so they are never cancelled.
+      group: ${{ github.event.release.prerelease && 'pre-release' || github.event.release.tag_name || inputs.tag }}
+      cancel-in-progress: true
```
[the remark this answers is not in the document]
cdunster: It's fine. If someone if manually triggering a rebuild of all the scenarios on a temporary tag such as `main-<sha>` then something unusual is happening and we probably don't want to cancel it anyway so we should just let it run.

PR #619, opened 2026-04-20
Only build changed scenarios based on Nix input hash

cdunster:
### Option 2 of 2

> [!Note]
> This is an alternative to #613.

> [!Warning]
> Only one of this or #613 should be merged.

### Summary

This is an alternative to #613 but whereas that one triggers rebuilds based on what files changed, this PR instead calculates the Nix input derivation hashes of each scenario and compares it to the one stored for the latest release and rebuilds scenarios where this differs.

This means that the hash needs to be stored with the release and then recalculated in every PR to check for changes but it should be more robust as it should detect any changes instead of just the ones specified.

### TODO:


cdunster:
Closing in favour of #613.

PR #656, opened 2026-07-02
fix(threefold): skip opted-out nodes during allocation

### LINKED ISSUE #649: Investigate Threefold node allocation failures
[https://github.com/holochain/wind-tunnel/actions/runs/27387605029/job/80937981357](<https://github.com/holochain/wind-tunnel/actions/runs/27387605029/job/80937981357>)

AC:

- [X] Identify the issue with Threefold integration, and if feasible within scope of this issue then fix
- [X] (Assuming first AC success) Estimate remaining credit on Threefold account
- [X] (Assuming first AC success) Measure one or more runs once working and estimate remaining weeks/months of use with existing credit

veeso:
### Summary

Fixes the ThreeFold scaled workflow so it skips nodes that TFChain would reject during contract creation. The workflow now also records the wallet balance before deploy and after cleanup, then reports the estimated cost and remaining weekly or monthly usage in the GitHub Actions summary.

Closes #649

### TODO:

Notes: this does not add or change a scenario. The unchecked Nomad item needs the canonical scaled workflow to run after this branch is available on GitHub.

Manual validation note: the canonical scaled workflow run on 2026-07-03 confirms the original issue is fixed: V3 billing opted-out ThreeFold nodes are skipped before contract creation. The run still fails because current usable ThreeFold capacity is below the scenario demand: it requested 238 VMs and found 203 candidate slots after filtering. Even without the 20-node buffer it would need 218 slots, so this is a capacity limitation rather than the original allocation bug.

### DIFF
--- a/threefold/deployer/main.go
+++ b/threefold/deployer/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"math"
+	"math/big"
 	"net"
 	"os"
 	"regexp"
@@ -14,6 +15,8 @@ import (
 	"strings"
 	"sync"
 
+	gsrpcTypes "github.com/centrifuge/go-substrate-rpc-client/v4/types"
+	substrate "github.com/threefoldtech/tfchain/clients/tfchain-client-go"
 	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/deployer"
 	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/workloads"
 	"github.com/threefoldtech/tfgrid-sdk-go/grid-client/zos"
@@ -37,6 +40,8 @@ const (
 	deploymentNameSuffix = "_dl_"
 	vmNameSuffix         = "_vm_"
 	networkNameSuffix    = "_net_"
+	githubOutputFormat   = "github"
+	textOutputFormat     = "text"
 )
 
 var prefixPattern = regexp.MustCompile(`^[A-Za-z0-9_]+$`)
@@ -53,6 +58,21 @@ type deployConfig struct {
 	prefix         string
 	maxConcurrency int
 	regionOrder    []string
+}
+
+type accountSnapshotConfig struct {
+	mnemonic string
+	network  string
+	format   string
+}
+
+type accountSnapshot struct {
+	address       string
+	freeTFT       string
+	reservedTFT   string
+	miscFrozenTFT string
+	freeFrozenTFT string
+	spendableTFT  string
 }
 
 type nodePlacement struct {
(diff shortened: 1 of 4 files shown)


.github/zizmor.yml
ThetaSinner, line 1: This is appearing in multiple repos but I'm not aware that it's been discussed anywhere. Normally we do discuss new tooling. It's not a workflow and it's not part of GitHub's built in tooling as far as I can see from its docs. So why does this file need to be checked in under `.github`?

Can you make a case for this new tooling please and we can discuss how we get it integrated across repos in their recommended way?

threefold/deployer/main.go
```diff
@@ -459,3 +559,174 @@ func capacityForNode(n types.Node, cfg deployConfig) (int, int64, int64, int64)
...
+		storageGetter: storageGetter,
+		accountID:     accountID,
+	}
+
+	checker.isAllowedTwinAdmin, err = checker.fetchAllowedTwinAdmin()
+	if err != nil {
+		return nil, fmt.Errorf("check allowed twin admin status: %w", err)
+	}
+	if checker.isAllowedTwinAdmin {
+		log.Printf("deployer wallet is an allowed twin admin; V3 billing opted-out ThreeFold nodes are allowed")
```
ThetaSinner, line 592: I'm not clear what's going on here. Is it possible that our walled isn't allowed to be used with the v3 network? That's the problem I was pointing to in discussion yesterday morning. What is this if/else currently reporting?

.github/zizmor.yml
veeso: Yes, I personally use it, and I forgot to put it into ignore list.

> The reason it appears sometimes is that my skills for working on github workflows enforce linting with zizmor which wants hash instead of tags by default, so then I add the rules file, but I forgot to ignore it.

threefold/deployer/main.go
veeso: It's not fully clear to me, and I have a limited understanding of. This is what Codex told me about:


> Physically, enabling it is a chain/admin operation, not a wind-tunnel code fix.
>
>  The path would be:

>  1. Create a dedicated deployer wallet for this purpose, not the normal CI wallet.
>  2. Make sure that wallet has a TFChain twin/account and enough TFT.
>  3. Ask the ThreeFold council/admin process to call:
>
>  TfgridModule.add_twin_admin(<that wallet account>)
>
>  That puts the wallet in AllowedTwinAdmins.
>
>  4. Put that dedicated wallet mnemonic in CI.
>  5. Our deployer already checks AllowedTwinAdmins; once the wallet is listed, opted-out nodes would no longer be filtered.
>
>  The danger is real: that wallet can deploy on all opted-out v3 nodes, which are not normal paid public capacity. They are migration/admin capacity. If the CI secret leaks or a workflow misfires, we could use nodes that farmers intentionally removed from normal v3 use.

So honestly, the behaviour seems correct. I would not make any action without asking an admin honestly.

I have added a better comment to say what that is reporting.

ThetaSinner (approved):

ThetaSinner:
Sorry I was slow getting to this, I kept getting sidetracked with tasks today and only just catching up on my task list

PR #661, opened 2026-07-17
feat: Update to Holochain 0.7.0-rc.1

ThetaSinner:
### TODO:

- [ ] If this PR adds a new scenario, I have copied the [new scenario checklist](https://github.com/holochain/wind-tunnel/tree/main/docs/new-scenario-checklist.md) into this PR description and ticked off each applicable item

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,7 +41,11 @@
         git-hooks.flakeModule
       ];
 
-      systems = builtins.attrNames inputs.holonix.devShells;
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "aarch64-darwin"
+      ];
 
       perSystem =
         { inputs'
--- a/flake.nix
+++ b/flake.nix
@@ -4,8 +4,6 @@
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-26.05";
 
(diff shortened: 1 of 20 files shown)


flake.nix
```diff
@@ -33,15 +31,18 @@
...
       imports = [
         git-hooks.flakeModule
       ];
 
-      systems = builtins.attrNames inputs.holonix.devShells;
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "aarch64-darwin"
+      ];
```
cdunster, line 45: I feel like we should just leave this as the Holonix systems and then remove the darwin-x86 support from Holonix instead of changing it everywhere?

cdunster (commented):
The most important thing is from the Bot; the default Holochain binaries in the Nomad workflows should be updated.

flake.nix
ThetaSinner: I'm thinking that because darwin-x86 is still working in Holonix CI, that it's quite nice to keep it there for people to use even if it's being gradually deprecated. 

I definitely agree it'd be the tidier solution to do that but I'm having trouble committing to its removal :D It's going from everywhere except binaries, kangaroo-electron and Holonix - so basically we don't test for it any more but it's still available to users if they want it.

PR #667, opened 2026-07-23
chore(threefold): decreased proportionally the threefold scenarios required nodes to not exceed 150

### LINKED ISSUE #657: Avoid over-provisioning Threefold VMs for serial canonical scaled runs
## Description

Canonical scaled Threefold runs are still failing before the scenarios start because there is not enough Threefold capacity for the requested VM count.

The observed failure is:

```text
insufficient preflight capacity: requested 225 VMs, estimated 185 slots across 43 candidate nodes (short by 40)
```

The problem is that we are actually requesting so many nodes in the scenarios; see <https://github.com/holochain/wind-tunnel/blob/main/nomad/job-variants/canonical-scaled/vars/zero_arc_create_data.json>.

We must reduce the amount while maintaining balance, aware that this could break again if the available VMs drop.

## Acceptance criteria

- [ ] Canonical scaled Threefold runs do not fail preflight because the workflow asks for more VMs than are needed for the scenario that is about to run.

veeso:
## Summary

reduced the amount of requested nodes for threefold scenarios in order not to exceed 150.

closes #657

### TODO:

### DIFF
--- a/nomad/job-variants/canonical-scaled/vars/dht_sync_lag.json
+++ b/nomad/job-variants/canonical-scaled/vars/dht_sync_lag.json
@@ -6,11 +6,11 @@
   "assignments": [
     {
       "behaviour": "write",
-      "nodes": 212
+      "nodes": 127
     },
     {
       "behaviour": "record_lag",
-      "nodes": 38
+      "nodes": 23
     }
   ]
 }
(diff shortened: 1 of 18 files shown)


veeso:
It failed to find most of runners while waiting, but now it runs. https://github.com/holochain/wind-tunnel/actions/runs/29988679720/job/89222219297

ThetaSinner (approved):
Great, there are more changes we could make and we could push closer to the limit of nodes but I think this is a sensible, conservative change that we can build on top of if it doesn't solve the problem 👍🏻

PR #708, opened 2026-08-27
feat(peerkit)!: turn first-connection scenario into hole punch & throughput test

### LINKED ISSUE #692: Extend Peerkit scenario to a hole punch test
The existing scenario just tests hole punching. Extend it to
- connect to a random peer
- disconnect from them

Some metrics that must be included in the scenario:
- Peer discovery time
- Connection time
- Type of established connection (direct/relayed)
- Number of errors

veeso:
closes #692


CHANGELOG.md
```diff
@@ -9,12 +9,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 
 - Add `peerkit_client_instrumented` and `peerkit_wind_tunnel_runner` crates, providing a Peerkit binding for wind-tunnel alongside the existing Holochain and Kitsune bindings.
-- Add `peerkit_first_connection` scenario, exercising a first connection between two Peerkit peers via a deployed relay.
+- Add `peerkit_hole_punch` scenario, exercising Peerkit's hole punching and throughput via a deployed relay. Every node connects to up to `PEERKIT_MAX_PEERS` discovered peers at a time, sends `PEERKIT_MESSAGES_PER_PEER` messages of `PEERKIT_MESSAGE_BYTES` bytes to each, and disconnects, repeating on a `PEERKIT_CYCLE_INTERVAL_MS` interval. Peer identities are random per agent, so any number of agents may share the single `node` behaviour. Metrics record peer discovery time, connection type (direct/relayed), and send/receive batch durations and byte counts, alongside an error count by kind.
```
jost-s, line 12:
```suggestion
- Add `peerkit_hole_punch` scenario, exercising Peerkit's hole punching and throughput via a deployed relay.
```

nomad/job-variants/canonical/vars/peerkit_hole_punch.json
```diff
@@ -0,0 +1,19 @@
...
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" }
```
jost-s, line 17:
```suggestion
    {
      "behaviour": "node",
      "nodes": 10
    }
```

nomad/job-variants/demo/vars/peerkit_hole_punch.json
```diff
@@ -0,0 +1,11 @@
+{
+  "job_name": "peerkit_hole_punch_demo",
+  "scenario_name": "peerkit_hole_punch",
+  "runtime": "peerkit",
+  "duration": 300,
+  "assignments": [
+    { "behaviour": "node" },
+    { "behaviour": "node" },
+    { "behaviour": "node" }
```
jost-s, line 9:
```suggestion
    { "behaviour": "node", "nodes": 3 }
```

framework/runner/src/run.rs
```diff
@@ -217,6 +217,18 @@ pub fn run<RV: UserValuesConstraint, V: UserValuesConstraint>(
...
+                                        log::debug!("Stopping agent {agent_name}");
+                                        break;
+                                    }
+
+                                    std::thread::sleep(Duration::from_secs(1));
+
+                                    if cycle_shutdown_receiver.should_shutdown() {
+                                        log::debug!("Stopping agent {agent_name}");
+                                        break;
+                                    }
```
jost-s, line 231: What's the rationale for this change?

bindings/peerkit_runner/src/common.rs
```diff
@@ -209,21 +245,28 @@ mod tests {
...
+        use std::os::unix::fs::PermissionsExt as _;
+
+        let (seed, agent_id) = generate_identity();
+        let path = write_identity_file(&agent_id, &seed).unwrap();
+        let mode = std::fs::metadata(&path).unwrap().permissions().mode();
+        std::fs::remove_file(&path).unwrap();
+
+        assert_eq!(mode & 0o777, 0o600);
+    }
+
```
jost-s, line 269: What is this test for? The scenarios use throwaway identities anyway, so it's not a risk.

bindings/peerkit_runner/src/common.rs
```diff
@@ -178,23 +172,65 @@ pub fn send_text(
...
+    if let Some(identity_path) = ctx.get_mut().identity_path.take()
+        && let Err(e) = std::fs::remove_file(&identity_path)
+    {
+        log::warn!(
+            "failed to remove identity key {path}: {e}",
+            path = identity_path.display()
+        );
     }
-    Ok(())
+    shutdown
```
jost-s, line 233: Equally this security isn't required.

bindings/peerkit_runner/src/common.rs
```diff
@@ -37,84 +35,65 @@ pub fn to_connection_string(relay_dial_addrs: Vec<String>) -> String {
...
-    behaviour: &str,
-) -> String {
-    derive_identity(ctx.runner_context().get_run_id(), behaviour).1
-}
+/// Write an agent's private key seed to a file that only its owner can read.
+///
+/// The file is created with its restrictive mode already applied rather than
+/// narrowed afterwards, so the key is never readable by other users, not even
+/// for the instant between creation and the permission change.
+fn write_identity_file(agent_id: &str, seed: &[u8; 32]) -> anyhow::Result<PathBuf> {
```
jost-s, line 57: The increased security here is also not necessary. All just throw-away identities.

.github/workflows/test.yaml
```diff
@@ -258,17 +260,31 @@ jobs:
...
+          # Exiting 0 is not enough: without at least one connection the
+          # scenario would idle through discovery failures and still pass.
+          grep -q 'wt.custom.peerkit_connection_established' "$scenario_log" \
+            || { echo "no peer connection was established during the run"; exit 1; }
+
+          # A connection on its own does not prove the nodes exchanged
+          # anything, so require a dispatched message batch too.
+          grep -q 'wt.custom.peerkit_send_batch' "$scenario_log" \
+            || { echo "no message batch was sent during the run"; exit 1; }
+
```
jost-s, line 287: This is not common. We're judging the scenarios' correctness based on metrics.

bindings/peerkit_client/src/node.rs
```diff
@@ -22,30 +24,175 @@ pub struct PeerkitNodeConfig {
...
+    /// First 64 characters of the message text (enough for scenario headers;
+    /// full payloads are not retained to bound memory).
+    pub text_prefix: String,
+    /// Full byte length of the message text.
+    pub len: usize,
+    /// When the message line was read from the CLI's stdout.
+    pub received_at: Instant,
+}
+
+/// Snapshot of one row of the `peers` command output.
```
jost-s, line 41:
```suggestion
/// A peer's agent info as returned by the CLI
```

bindings/peerkit_client/src/node.rs
```diff
@@ -22,30 +24,175 @@ pub struct PeerkitNodeConfig {
...
+    /// lock, preventing late send failures from being mistaken for command
+    /// responses.
+    active_command: Arc<AtomicU8>,
+    /// Number of completed commands when the active command started. This
+    /// prevents a delayed event after the command's completion prompt from
+    /// being mistaken for that command's response.
+    command_start_completed: Arc<AtomicU64>,
+    /// The stdout reader is joined during shutdown so all buffered output is
+    /// applied to state before teardown drains it.
+    reader: Mutex<Option<tokio::task::JoinHandle<()>>>,
```
jost-s, line 172: This looks like it's tracking which command is active and other things. What is this for?

scenarios/peerkit_hole_punch/README.md
```diff
@@ -0,0 +1,146 @@
...
+3. Send `PEERKIT_MESSAGES_PER_PEER` messages of `PEERKIT_MESSAGE_BYTES` bytes
+   each to every peer connected this cycle.
+4. Fold any messages received from other agents' batches into per-sender
+   trackers, emitting a metric once a batch of `PEERKIT_MESSAGES_PER_PEER`
+   messages from one sender has fully arrived. This is repeated over a short
+   grace period because local send completion includes stream backpressure but
+   does not prove remote application delivery; hanging up straight after the
+   last dispatch can still cut a batch short on the receiving peer.
+5. Disconnect from every peer connected this cycle.
+6. Sleep for `PEERKIT_CYCLE_INTERVAL_MS` before starting the next cycle.
```
jost-s, line 22: The scenario is more complicated than I thought. I can see how it's difficult to make this a holepunch _and_ throughput test at the same time. Peers are waiting for all incoming messages to have arrived before disconnecting.

The scenario could instead be more aggressive about disconnecting and not heed whether a full batch has been received. That would reduce the logic to keep track of how many messages per peer have been received.

Would it be better to keep the scenarios separate, so that this is a pure holepunch scenario? Is the throughput really that interesting? Maybe I was mistaken there. What I'm really curious about is the holepunch rate.

@mattyg @cdunster @veeso Requesting your input.

mattyg: Yes, I think that we don't want to confuse what we're measuring here.

My feeling is that in general for wind tunnel scenarios we want a lot of small scenarios that let us measure a single thing, and then after those are done, *maybe* consider a few higher-level "smoke" scenarios that use randomness to simulate real world behavior exercising lots of different functionality.

So for measuring hole punching, I think the simplest scenario is just connect to a single peer, send a message, receive a message, disconnect, repeat.

mattyg:
Is this supposed to resolve #692? It's not in the description to auto-close it.

scenarios/peerkit_hole_punch/README.md
jost-s: I agree, better to keep this simple. We can decide about additional scenarios, but this should be just a holepunch scenario.

I've updated the issue. When establishing a connection, a stream is opened for the handshake. Maybe that's even enough, just open a connection to a random peer, wait for the direct connection upgrade on a timeout and close the connection?

.github/workflows/test.yaml
veeso: Removed the metric-grep assertions and in-memory reporter from the smoke test. The step now relies on scenario metrics for correctness while set -e still ensures a nonzero scenario exit fails CI.

bindings/peerkit_client/src/node.rs
veeso: This state attributes output to the one serialized REPL command in flight. Successful Peerkit commands, including send, are acknowledged only by the next prompt, while asynchronous events also redraw the prompt. Tracking the active command and its completion baseline prevents an event prompt or a delayed response from satisfying the wrong command; after cancellation or timeout the node is marked desynchronized so a late prompt cannot be attributed to the next command.

bindings/peerkit_runner/src/common.rs
veeso: Removed the identity-file permission test; these identities are throwaway.

CHANGELOG.md
veeso: Applied: the changelog entry is now the concise single-line scenario description.

framework/runner/src/run.rs
veeso: The one-second delay prevents persistently failing behaviours from hot-looping. In this scenario, a timed-out or cancelled REPL command intentionally leaves the client permanently desynchronized, so every later iteration fails immediately. The delay bounds CPU and log volume, while the checks before and after it preserve responsive shutdown. The regression test covers the retry rate.

bindings/peerkit_runner/src/common.rs
veeso: Removed teardown-time key cleanup and the context state used only to support it.

bindings/peerkit_client/src/node.rs
veeso: Updated the documentation to describe this as a snapshot of one peers-command row, with terminal punctuation.

nomad/job-variants/canonical/vars/peerkit_hole_punch.json
veeso: Applied: the canonical assignment now uses one node entry with nodes set to 10.

bindings/peerkit_runner/src/common.rs
veeso: Simplified identity-file creation to a normal temporary-file write; the restrictive create-mode handling was unnecessary for throwaway identities.

nomad/job-variants/demo/vars/peerkit_hole_punch.json
veeso: Applied: the demo assignment now uses one node entry with nodes set to 3.

scenarios/peerkit_hole_punch/README.md
veeso: @jost-s @mattyg refactored to just punch hole

bindings/peerkit_client/src/stream.rs
mattyg, line 1: It looks like peerkit can currently only be consumed as a cli or library. But here we have a case where we require an api or driver interface. I would expect other users of peerkit might also need a similar interface (any non-js consumers). Does it make sense to make a real driver, owned by the peerkit project and repo, and then wind-tunnel can just consume it?

Parsing text via a repl to wrap the cli feels like an overly complicated and brittle approach that only makes sense assuming these limitations that maybe should be addressed directly.

That said if we want to get this in now and refactor after, that's fine with me.

scenarios/peerkit_hole_punch/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+    relayed_seen: bool,
+    peer_present: bool,
+    status: Option<PeerStatus>,
+) -> Option<ConnectionOutcome> {
+    match status {
+        Some(PeerStatus::Direct) => Some(ConnectionOutcome::Direct),
+        Some(PeerStatus::Relayed) => None,
+        Some(PeerStatus::NotConnected) if relayed_seen => Some(ConnectionOutcome::Lost),
+        Some(PeerStatus::NotConnected) => None,
+        None if !peer_present && relayed_seen => Some(ConnectionOutcome::Lost),
```
mattyg, line 108: Should we distinguish `NeverConnected` from `Lost`? Or I might be confused about the difference between this case and the Lost case above.

mattyg:
Actually I see this doesn't include the summariser and visualiser changes, so maybe it shouldn't close #692, or we should create a follow-up issue for those, or we should include those here.

scenarios/peerkit_hole_punch/src/main.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+}
+
+fn node_cycle(
+    ctx: &mut AgentContext<PeerkitRunnerContext, PeerkitAgentContext>,
+) -> anyhow::Result<()> {
+    let upgrade_timeout =
+        Duration::from_millis(env_u64("PEERKIT_DIRECT_UPGRADE_TIMEOUT_MS", 10_000)?);
+
+    report_discovery_times(ctx)?;
```
mattyg, line 173: If nodes are discovered during the final loop then this won't get called and the data wont get reported. Should we call it once more in teardown?

bindings/peerkit_client/src/stream.rs
veeso: Worth discussing with the team, but for me it's out of scope atm

veeso:
@mattyg summariser is implemented in a separate task https://linear.app/hcf/issue/HC-876/add-summariser-report-for-peerkit-hole-punch-scenario

scenarios/peerkit_hole_punch/src/main.rs
veeso: fixed

scenarios/peerkit_hole_punch/src/main.rs
veeso: fixed by distinguishing two states

bindings/peerkit_client/src/stream.rs
jost-s: We could add a driver, although Peerkit is written in TS with general accessibility in mind. The wrapper could be moved to the Peerkit repo.

scenarios/peerkit_hole_punch/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+    if let Err(error) = disconnected {
+        if ctx.shutdown_listener().should_shutdown() {
+            return Ok(());
+        }
+        log::warn!("disconnect from {alias} failed: {error:#}");
+        report_error(ctx, "disconnect");
+    }
+
+    sleep_interval(ctx)
```
jost-s: Why this sleep? I've included sleeps in some scenarios in the past but was informed that Wind Tunnel is a performance test framework, and behaviors should not use sleeps unless they malfunction completely otherwise. If performance degrades, that's something that should surface through metrics.

jost-s (commented):
Looks good to me except the sleep in the behavior

scenarios/peerkit_hole_punch/src/main.rs
veeso: Removed. Also added instructions for agents to flag those.

bindings/peerkit_client/src/node.rs
jost-s: Also something that should be made more machine-consumer friendly in peerkit. I've not paid attention to that, I just looked at the UX as a human.

jost-s (approved):

Issue #690, opened 2026-08-18
Flaky test on macos `tests::say_something_to_other_chatter`


ThetaSinner:
e.g. https://github.com/holochain/wind-tunnel/actions/runs/32040655182/job/95419248013

PR #716, opened 2026-09-02
build: pin kitsune2 simultaneous-open fix

veeso:
### Summary

Pins the Kitsune dependencies to commit `5e96fc82bed1b60b5c97ee20475e3fc716bd4dd5`. That commit waits for two peers to settle on one connection before sending their first gossip messages, which prevents the macOS failure tracked in #690. The unchanged failing test passed 10 consecutive times locally on macOS.

### TODO:

- [x] All code changes are reflected in docs, including module-level docs. No Wind Tunnel code or public behavior documentation changed.
- [x] All new/edited/removed scenarios are reflected in the summary visualiser tool. No scenarios changed.
- [ ] I ran the Nomad CI workflow successfully on my branch. It is not required for this dependency-only test fix.
- [x] If this PR adds a new scenario, I have copied the new scenario checklist into this PR description. This PR does not add a scenario.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1919,8 +1919,8 @@ dependencies = [
  "happ_builder",
  "holochain_types",
  "holochain_wind_tunnel_runner",
- "kitsune2_api",
- "kitsune2_core",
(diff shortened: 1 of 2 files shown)


ThetaSinner (changes requested):
Can we use a release of Kitsune2 here please, not a Git revision?

PR #724, opened 2026-09-16
feat/712 add smoke test argument to fail scenarios on panic

### LINKED ISSUE #712: Fail scenarios when an agent thread panics
## Description

A scenario runs each agent in a separate thread. When an agent thread panics, the runner logs that the thread could not be joined, but it continues and returns a successful exit status. This can make a scenario and its smoke test appear successful even though the workload did not run.

We verified this with the existing app_install smoke test. With WT_METRICS_DIR absent, both agent threads panicked with "WT_METRICS_DIR must be set", the runner logged two failed thread joins, and the command still exited with status 0. The conductor_startup scenario exposed this because it has its own completion validation after the shared runner returns.

The shared runner must report a failure whenever any agent thread cannot be joined. This should apply to every scenario rather than requiring scenario-specific completion checks.

## Notes

The failed thread join is currently only logged in framework/runner/src/run.rs. The behaviour was observed in https://github.com/holochain/wind-tunnel/actions/runs/33496374850/job/99819566914?pr=711.

## Acceptance criteria

- [ ] Ensure that if `WT_METRICS_DIR` is missing, when influx file is set as metrics collection system, the scenario fails with an error bail.
- [ ] At `run.rs` put a guard with a CLI option (if possible) or alternatively with an env variable **that must be set** for smoke test, to panic on scenario panics.

   ```rust
       if let Err(e) = handle.join() {
            if args.smoke_test /* std::env::get("SMOKE_TEST_...").is_some() */ {
                anyhow::bail!("Could not join thread for test agent {index}: {e:?}")
            else {
               log::error!("Could not join thread for test agent {index}: {e:?}")
            }
        }
   ```

- [ ] Ensure the argument / env variable is set for smoke test workflow.

veeso:
### Summary

Add an argument to scenario runner to fail in case of agent panic.

This flag has also been added to all the smoke tests. It won't affect nomad runs.

`WT_METRICS_DIR` flag presence has been enforced in case `influx-file` is passed as metrics collection option.

### kitsune_continuous_flow fix
Agent-owned async resources now shut down within Tokio, preventing `kitsune_continuous_flow` from panicking after a successful run. (see https://github.com/holochain/wind-tunnel/actions/runs/35105557926)

### TODO:


framework/runner/tests/hook_error_handling.rs
```diff
@@ -25,11 +38,106 @@ fn sample_cli_cfg() -> WindTunnelScenarioCli {
...
         soak: false,
         no_progress: true,
+        fail_on_agent_panic: false,
         reporter: ReporterOpt::Noop,
         run_id: None,
     }
 }
 
+#[test]
+fn accepts_fail_on_agent_panic_flag() {
```
mattyg, line 48: doesn't seem necessary

veeso: legit, removed

mattyg (approved):

